PR #34, opened 2015-01-06
Integrating react into code base.

Gozala:
This is just re-targets existing pull request that was made here https://github.com/paulrouget/firefox.html/pull/94 which was synchronized with master up until migration commit.


paulrouget:
Alright, let's see if that sticks :)

PR #44, opened 2015-01-09
Add popups when errors occur, fix #38

### LINKED ISSUE #38: CertErrors should use about:netError
STR:
- Load https://linuxfr.org
- empty page

PalmerAL:
Add popups for the following errors alerting the user as to what went wrong (fixing #38):
-no internet connection
-redirect loop
-corrupted content
-remote xul
-invalid certificate

(eventually, the other error types should have popups as well, but for right now it is just these 5)

some questions:
-what should the popup messages be? Right now most of the popups are just a basic explanation or copied from firefox, but this is probably not enough.
-should there be one global popup at the document root (like there is now) or should there be individual popups attached to the tab-iframes?
-Is the trigger on the invalid certificate too broad? Right now, it just triggers on anything with type "other", but are there other situations where this type occurs that need to be taken into account?
screen shot 2015-01-09 at 5 02 06 pm

### DIFF
--- a/apps/browser/css/layout.css
+++ b/apps/browser/css/layout.css
@@ -42,6 +42,11 @@ tab-iframe, box, hbox, vbox, spacer {
   white-space: pre;
 }
 
+tab-iframe {
+  overflow: hidden;
+  position: relative;
+}
+
 hbox {
   flex-flow: row;
 }
--- a/apps/browser/css/popup.css
+++ b/apps/browser/css/popup.css
@@ -1,0 +1,52 @@
+.popup {
+  position:fixed;
+  top:30vh;
+  left:30vw;
+  background:#F8F8F8;
+  color:#777;
+  padding:2em;
+  box-shadow:0px 27px 24px 0px rgba(0, 0, 0, 0.2), 0px 40px 77px 0px rgba(0, 0, 0, 0.22), 0px -5px 24px 0px rgba(0, 0, 0, 0.2);
+  min-width: 200px;
+  min-height: 40vh;
+  max-height: 70vh;
+  overflow: auto;
+  width:40vw;
+  height:auto;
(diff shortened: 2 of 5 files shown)


jdm:
:+1: That's pretty slick.

paulrouget:
Looks nice. Does it need to be a popup? Can't it be a page in the tab?

PalmerAL:
It probably could. That would require putting it in tab-frames, which would be somewhat more complicated, but it shouldn't be too much work.

PalmerAL:
At some point, though, I'd like to have <code>alert()</code> s from webpages to use html popups in the app, and having this be a page in the tab would not work well for that.

paulrouget:
I'd prefer this code to be at the app level. Not in the runtime.

And yes, we will need dialog management, like in Gaia system app. But for network errors & co, it should stay as a page.

PalmerAL:
Should this be an element that looks like a page and takes up the whole-tab iframe element, or an actual page displayed in the iframe that is located as its own file in the app?

PalmerAL:
Updated so that you can make the popups in-page by adding two additional keys to the options: <code> page </code> and <code>insertiontarget</code>. The okay button now affects all popups that are currently open, not just the one in which the popup is clicked.

paulrouget:
Why not a page loaded in the iframe?

PalmerAL:
In the new runtime, the app:// protocol doesn't work anymore, so unless you put this in the runtime there doesn't appear to be a way to do this.

paulrouget:
Oh. I didn't know we broke that. This should be fixed as soon as we migrate to the B2G runtime. In the meantime, you can still use HTMLRunner.

PalmerAL:
I'll close this for now... Will come back to after the migration to B2G is complete.

PR #49, opened 2015-01-11
add a new tab button

PalmerAL:
screen shot 2015-01-11 at 10 52 44 am


paulrouget:
This button should be in the tabstrip, not the navbar.

paulrouget:
Irrelevant now.

Issue #39, opened 2015-01-07
ReactJS feedback


paulrouget:
I backedout React integration.

I played with it today. This new code base feels over engineered for such a simple UI (4 buttons, 2 inputs, tabs), and we still don't know what the UI will look like in the future.

I am not ruling out React, but it's a premature optimization. Let's reconsider using React later.

andreasgal:
I love this. Practical applied feedback. Tried it early. Didn't work great. Undo. Try something else or try again later. This is priceless learning experience.

Gozala:
It would have being premature optimization if intent of using react was optimization. While indeed React is bundled with a lot of optimizations that was not the main point (not to me anyway).

I've listed bunch of reasons in original pull request https://github.com/paulrouget/firefox.html/pull/94

In my experience (specially at Mozilla) simple UIs quickly grow into complex ones and code quickly becomes unmaintainable hairball of state distributed across different components that are unnecessarily coupled. Usually that means making any change without breaking anything else is quite difficult.

The reason I wanted to integrate React early on was to avoid same route. React does provide & enforces very clear separation of concerns between components & localizes state into single place. This combined with unidirectional data flow model it encourages makes it a lot easier to reason about code & make changes without breaking unrelated parts. It is true that benefits become more apparent as code grows, but also does cost of switching (it already took me few days month from now it will be too much work to be worth it).

paulrouget:
> In my experience (specially at Mozilla) simple UIs quickly grow into complex ones and code quickly becomes unmaintainable hairball of state distributed across different components that are unnecessarily coupled

I agree. So let's make sure it doesn't happen.

I'm absolutely certain that the current code will be rewritten many times. There's no value in the current code base. It's just a sandbox.

We will revisit that later.

paulrouget:
@Gozala, can you explain `virtual-dom.js`, `links.js`, `component.js` and `element.js`?

Gozala:
@paulrouget 

- `virtual-dom.js` Essentially it's just exports `React.DOM` as `html`, which is used in `render` methods to describe the node associated with a component. React takes care of DOM updates for you. In terms of API I think it's pretty straight forward `html.div({ className: "bla", id: "foo" }, children)`. 

[...]

PR #51, opened 2015-01-12
Simplify code layout.

Gozala:
This is a patch we've talked about before.


paulrouget:
I merged your branch, and I get:
```
module.js:340
    throw err;
          ^
Error: Cannot find module '/Users/paul/github/browser.html/node_modules/firebox/firebox.js'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:906:3
```

paulrouget:
Can you work directly on the `mozilla/react` branch?

paulrouget:
And I'm not sure yet we want to keep a one single app layout.

bbondy:
For the React branch could we:
i) Use JSX (w/ harmony flag from the compiler) directly for the files that will use React, we could scope it to just the files with React in them. You can have it watch changes w/ jsx --harmony -w js/ build/
ii) Remove any abstractions and just use React directly.

Gozala:
> I merged your branch, and I get:

You have to run `npm install` to install firebox dependency.

Gozala:
> And I'm not sure yet we want to keep a one single app layout.

Can we can keep things simple for now and if we end up with more complex scenario than we reflect that. It seems awkward to have complex layout as we may go with something more complex in a future.

Gozala:
> For the React branch could we:
i) Use JSX (w/ harmony flag from the compiler) directly for the files that will use React, we could scope it to just the files with React in them. You can have it watch changes w/ jsx --harmony -w js/ build/

I do not think JSX brings enough value to the table to justify compilation step, not to say that in that case we be even writing standard JS. 

> ii) Remove any abstractions and just use React directly.

There are couple of reasons why those abstraction was introduced:

1. React has bunch of limitations (some where mentioned earlier in this thread) few abstractions here are specially created to workaround them.
2. Limitations of react do get into our way & react proved itself to be inflexible with non standard or new html stuff (like web components). For this reason we would like to keep react swappable for example with something like [vtree](https://github.com/Matt-Esch/vtree) + [vdom](https://github.com/Matt-Esch/vdom). Use of abstraction will let us swap out react with something else without having to rewrite anything but an abstraction layer.
3. React does encourage no local state, but it does not enforces it hard enough. Introduced abstraction exposes just a good part of react enforcing that further.
4. React can not run in a web worker, while [vtree](https://github.com/Matt-Esch/vtree) + [vdom](https://github.com/Matt-Esch/vdom) can so keeping abstraction over these two options is good as it would allow us to experiment with moving some parts of code into web worker.

Gozala:
BTW jsx could be addressed with template strings that are standard an already in js so we could have something like: jsx`<div style="color=${color}">${text}</div>` if we wanted to, but I have not spend time on this, partially because I don't find that much better comparing to `html.div({ style: {color} }, text)`

bbondy:
Just weighing in, but feel free to do as you wish.

> I do not think JSX brings enough value to the table to justify compilation step, not to say that in 
> that case we won't be even writing standard JS.

In my experience JSX code base requires much less maintenance as React develops as a library than non JSX code using React. Also it's easier and cleaner to write.

> There are couple of reasons why those abstraction was introduced:

IMHO it's for a react branch so the extra abstractions aren't worth the extra complexity. If this was the mainstream branch that others bought into then maybe.  I'm not convinced that the limitations are indeed important and that we should just be embracing the framework and what it provides instead of fighting it.

paulrouget:
Please continue the discussion in issue #39

PR #63, opened 2015-02-12
Update README.md

GabrielDuque:


Gozala:
What a hell is that some kind of new form of spam ?

PR #67, opened 2015-02-13
Add hardcoded color themes

gordonbrander:
I'm still getting used to React and the codebase, so please do review and suggest improvements.

- Add concept of "dark" and "light" themes.
  - Allows us to adjust input color and borders.
- Set font to helvetica, make site bold.
  - Fix icon font weight.
- Lighten color of domain
  - We don't have a medium font weight, so visually lighten bold text by
    increasing opacity.


gordonbrander:
Note that Travis is failing on 3rd party libs I did not touch. We should fix that.

css/theme.css
```diff
@@ -110,7 +110,7 @@ iframe {
   flex-shrink: 0;
   border-radius: 3px;
   line-height: 30px;
-  background-color: #EEEEEE;
+  background-color: rgba(0, 0, 0, 0.07);
```
paulrouget: Nit: no space in `rgba()` values.

node_modules/browser/navigation-panel.js
```diff
@@ -14,17 +14,27 @@ const {navigateTo, showTabStrip, focus} = require('./actions');
...
 const url = require('./util/url');
 
-const WindowControls = Component(options =>
-  DOM.div({className: 'windowctrls'}, [
+const WindowControls = Component(({webViewer}) => {
+  const theme = webViewer.get('theme'),
+        closeColor = theme.get('close'),
+        minColor = theme.get('min'),
+        maxColor = theme.get('max'),
+        glyphsShowing = theme.get('glyphsShowing');
```
paulrouget: Don't chain the const. One const per line (for consistency with other files).

node_modules/browser/theme.js
```diff
@@ -0,0 +1,71 @@
...
+  location: "#eee",
+  buttons: "rgba(0, 0, 0, 0.5)",
+  domain: "rgba(0, 0, 0, 0.65)",
+  title: "rgba(0, 0, 0, 0.5)",
+  background: "#fff"
+};
+exports.defaultTheme = defaultTheme;
+
+const hardcodedColors = {
+  // [foreground, background]
```
paulrouget: `// [foreground, background, isDark]`

Maybe instead of a boolean, you can use a const. `IS_DARK` / `IS_LIGHT`. Booleans are "unknown meat".

node_modules/browser/theme.js
```diff
@@ -0,0 +1,71 @@
...
+  "supertimeforce.com": ["#2EBCEC", "#051224", true]
+};
+
+const expandTheme = (foregroundColor, backgroundColor, isDark) => ({
+  isDark: isDark,
+  glyphsShowing: true,
+  close: foregroundColor,
+  min: foregroundColor,
+  max: foregroundColor,
+  progress: isDark ? "rgba(255, 255, 255, 0.2)" : "rgba(0, 0, 0, 0.2)",
```
paulrouget: Is this used? We'll use that in a follow up I guess.

css/theme.css
```diff
@@ -231,27 +242,25 @@ iframe {
 }
 
 .windowctrl {
+  background: url('sprite-glyph.png') -100px 0 no-repeat;
```
paulrouget: What is this?

node_modules/browser/web-viewer/actions.js
```diff
@@ -21,15 +22,29 @@ const initial = fromJS({
...
+// Open new tab
+const open = () => initial.merge({
+  id: uuid()
+});
 exports.open = open;
 
+// Load initial tab.
+// @FIXME there may be a better way to handle this case: a more flexible `open`
+// function perhaps?
+const load = (uri) => initial.merge({
```
paulrouget: If this is used only once, maybe it's better to keep this code inline.

node_modules/browser/web-viewer/actions.js
```diff
@@ -21,15 +22,29 @@ const initial = fromJS({
...
+  id: uuid(),
+  uri: uri,
+  isSelected: true,
+  isFocused: true
+});
+exports.load = load;
```
paulrouget: If this function is used only once, maybe it's better to keep this code inline.

css/theme.css
gordonbrander: I'm all for consistency, and don't mind changing this, but curious if you feel it applies to other CSS functions like `linear-gradient`?

node_modules/browser/theme.js
gordonbrander: Good advice. Note that these booleans are localized to the file. These arrays are expanded into a more detailed "theme object" by `expandTheme`.

css/theme.css
gordonbrander: Thanks for catching this mistake. I had integrated glyphs for the stoplights, but pulled it out. The OSX default glyphs are too visually indistinct. I suspect that we might do better by tweaking the native style just a bit. Need some time to sketch this out.

node_modules/browser/web-viewer/actions.js
gordonbrander: I actually added this because we had a mysterious bug introduced by a hardcoded state object. I feel it's better to keep all the state initializers in one central, searchable place.

node_modules/browser/theme.js
gordonbrander: Let's talk about the progress bar. I think we can change the design just a bit to make it easier to color theme.

gordonbrander:
Looks like I'll need to rebase.

gordonbrander:
Ok, I merged changes on top of master, but did not flatten the commits. In my experience it causes more trouble than it is worth to flatten change history, but let me know if you feel strongly otherwise.

node_modules/browser/browser.js
```diff
@@ -98,6 +98,8 @@ const Browser = Component(options => {
 
   const input = options.cursor('input');
 
+  const backgroundColor = webViewer.getIn(["theme", "background"]);
```
Gozala: Nit: You can actually use just `webViewer.get(["theme", "background"])` although I don't know what the diff is to be honest.

node_modules/browser/navigation-panel.js
```diff
@@ -14,17 +14,27 @@ const {navigateTo, showTabStrip, focus} = require('./actions');
...
 const url = require('./util/url');
 
-const WindowControls = Component(options =>
-  DOM.div({className: 'windowctrls'}, [
+const WindowControls = Component(({webViewer}) => {
+  const theme = webViewer.get('theme');
+  const closeColor = theme.get('close');
+  const minColor = theme.get('min');
+  const maxColor = theme.get('max');
+  const glyphsShowing = theme.get('glyphsShowing');
```
Gozala: Nit: you could also do something this:

```js
const theme = webViewer.get('theme').toJSON();

// ...
style: { backgroundColor: theme.close}
```

or even:

```js
const {close, min, max, glyphsShowing} = webViewer.get('theme').toJSON();
```

node_modules/browser/navigation-panel.js
```diff
@@ -34,23 +44,32 @@ const WindowControls = Component(options =>
...
 
-const NavigationControls = Component(({input, webViewer, title}) =>
-  DOM.div({
+const NavigationControls = Component(({input, webViewer, title}) => {
+  const theme = webViewer.get('theme');
+  const buttonColor = theme.get('buttons');
+  const titleColor = theme.get('title');
+  const domainColor = theme.get('domain');
+
+  const location = webViewer.get('location')
```
Gozala: I would rather prefer to have something like:

```js
const theme = readTheme(webViewer);

return DOM.div({
     className: 'locationbar',
    onMouseEnter: event => showTabStrip(input),
}, [
     DOM.div({className: 'backbutton',
              style: theme.buttonColor,
// ...
```

node_modules/browser/web-viewer/actions.js
Gozala: I think I mentioned it in #77 as well, it would make sense to change open to:

```js
const open = (options={}) =>
  options.id ? initial.merge(options) : initial.merge(Object.assign(options, {id: uuid()}));
```

That way use of load(uri) will just be updated to `open({uri: 'http://...'})`

Gozala: Also the issues @gordonbrander mention is related to the fact that we start browser with one tab pre-opned and focused. Maybe starting blank is a better option ?

gordonbrander:
Given we just moved everything out of `node_modules, I think I'm just going to cherry pick these changes on to current master in a new branch.

PR #75, opened 2015-02-17
Move browser/ and shims/ out of node_modules

KevinGrandon:
Just a simple patch to move things into the js/ folder. This keeps the browser and shims folders intact.


browser.js
```diff
@@ -1,27 +1,27 @@
 require.config({
   scriptType: 'text/javascript;version=1.8',
-  baseUrl: 'node_modules/',
```
Gozala: I would prefer to keep `baseUrl` mapped to `node_modules` & instead just configure `browser` and `shims` paths, I think adding following to `paths` would do: 

```js
paths: {
  'browser': '../src/browser',
  'shims': '../src/shims',
  //...
}
```

Also how about calling directory `src` instead of `js`, just because most other projects tend to do that.

paulrouget:
Just did the migration. 35f161906c8bfcd72e1a100edd213a31c8246f8c. I'm not expecting to have *less* PR in the future. So let's do the dirty work now. And I probably broke travis.

PR #76, opened 2015-02-17
Fix jscs validation errors

KevinGrandon:
This is kind of funny. We've moved the JS into node_modules, but we're currently ignoring all style checks in the node_modules folder: https://github.com/mozilla/browser.html/blob/master/.jscsrc#L4

This fixes most of the style validation. We still ignore errors, but that will be fixed once we land #75 (or we could update the paths in .jscsrc).

### DIFF
--- a/node_modules/browser/actions.js
+++ b/node_modules/browser/actions.js
@@ -4,31 +4,31 @@
 
 define((require, exports, module) => {
 
-'use strict';
+  'use strict';
 
-const url = require('./util/url');
+  const url = require('./util/url');
 
-const makeSearchURL = input =>
-  `https://search.yahoo.com/search?p=${encodeURIComponent(input)}`;
-exports.makeSearchURL = makeSearchURL;
+  const makeSearchURL = input =>
+    `https://search.yahoo.com/search?p=${encodeURIComponent(input)}`;
+  exports.makeSearchURL = makeSearchURL;
 
-const readInputURL = input =>
-  url.isNotURL(input) ? makeSearchURL(input) :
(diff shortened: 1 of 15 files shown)


Gozala:
@KevinGrandon is there any way we could configure `jscs` to not require indentation for the wrapper function body ? I do plan to get rid of those wrappers soonish & would really like to avoid making complete reindent commits, that's why function wrapper body isn't indented in first place.

gordonbrander:
+1 I would prefer to fix the linter

KevinGrandon:
I don't really think it's worth it to be scared of indentation changes, there's not a ton of code :)

If we're not ready to migrate to another solution (not sure what you had in mind), then I'll see what options jscs has, I've never used it before. Switching to jshint might also work - jshint is more about syntax and less about style.

gordonbrander:
If we're going to do this, let's merge outstanding PRs that touch many files like https://github.com/mozilla/browser.html/pull/93 first. I've already re-written that one once with all the file reorganizations ;)

paulrouget:
Agreed with Kevin. Let's fix indentation in the code. Code base is pretty small today. And to be honest, the current indentation drives me nuts :)

paulrouget:
Sorry, I short-circuited you with PR #101

KevinGrandon:
> Sorry, I short-circuited you with PR #101

This is fine and well, but does make it a bit de-motivating to potential contributors. If you are looking for contributors for the project I suggest you stop doing that in the future. (I'll still do stuff if I have time)

PR #77, opened 2015-02-17
support onOpenWindow

paulrouget:
@Gozala For some reason, the built iframe is rendered as `<iframe hidden="">` instead of `<iframe hidden="true">`. Any idea why? `hidden=true` is normally used when `toggle` is used.

### DIFF
--- a/node_modules/browser/web-viewer.js
+++ b/node_modules/browser/web-viewer.js
@@ -12,7 +12,8 @@ const {isFocused} = require('./focusable');
 const {Element, Option, Field, Event} = require('./element');
 const Component = require('omniscient');
 const {Deck} = require('./deck');
-const {remove} = require('./deck/actions');
+const {open} = require('./web-viewer/actions');
+const {remove, append} = require('./deck/actions');
 
 const IFrame = Element('iframe', {
   isFocused: isFocused,
@@ -134,7 +135,12 @@ const WebViewer = Component(({item, items}) => {
   onFocus: WebViewer.onFocus(item),
   // onAsyncScroll: WebViewer.onUnhandled,
   onClose: event => remove(items, x => x.equals(item)),
-  onOpenWindow: WebViewer.onUnhandled,
+  onOpenWindow: event => {
+    items.update(items => {
+      const item = open({uri: event.detail.url});
+      return append(items, item);
+    });
+  },
   onContextMenu: WebViewer.onUnhandled,
   onError: event => console.error(event),
   onLoadStart: WebViewer.onLoadStart(item),
--- a/node_modules/browser/actions.js
+++ b/node_modules/browser/actions.js
@@ -19,7 +19,7 @@ const readInputURL = input =>
 exports.readInputURL = readInputURL;
 
 // Action takes state cursor for the web viwer and input location
-// and navigates that webViewr to that loctaion (if it's not valid
+// and navigates that webViewr to that location (if it's not valid
 // url either normalizes it or converts to search). Optional `focus`
 // can be passed as `false` to navigate to a url but not focus it.
 const navigateTo = ({input, webViewer}, location, focus=true) => {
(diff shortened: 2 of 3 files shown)


node_modules/browser/web-viewer.js
```diff
@@ -134,7 +135,13 @@ const WebViewer = Component(({item, items}) => {
   onFocus: WebViewer.onFocus(item),
   // onAsyncScroll: WebViewer.onUnhandled,
   onClose: event => remove(items, x => x.equals(item)),
-  onOpenWindow: WebViewer.onUnhandled,
+  onOpenWindow: event => {
+    items.update(items => {
+      const item = open().merge({uri: event.detail.url,
+                                 isSelected: false});
```
Gozala: Nit: I think we should move `isSelected: false` into [initial form](https://github.com/mozilla/browser.html/blob/master/node_modules/browser/web-viewer/actions.js#L12-L28) and we could also modify [`open`](https://github.com/mozilla/browser.html/blob/master/node_modules/browser/web-viewer/actions.js#L30) to let pass in additional config & avoid more allocations.

Gozala:
> For some reason, the built iframe is rendered as <iframe hidden=""> instead of <iframe hidden="true">. Any idea why? hidden=true is normally used when toggle is used.

I looked at that code and I'm under impression it has something to do with react itself. My understanding of global `hidden` attribute is that it's value does not matter, but rather it's presence or absence. I imagine what react does is updates value via `element.hidden = true` in which case result will be `hidden=""`. I confirmed that by running opening following url:

`data:text/html,<div>hi</div><script>document.querySelector("div").hidden = true; alert(document.querySelector("div").outerHTML);</script>`

I also believe that in some updates react does use `node.innerHTML = html` for optimisation purposes where I imagine generated `html` will have `hidden="true"` instead.

Anyway I think there is two ways we could go about it:


1. Handle `hidden` attribute on our own, as show with a patch onto of this pull below:

```diff
diff --git a/node_modules/browser/web-viewer.js b/node_modules/browser/web-viewer.js
index c6ae245..619cc50 100644
--- a/node_modules/browser/web-viewer.js
+++ b/node_modules/browser/web-viewer.js
@@ -9,7 +9,7 @@ define((require, exports, module) => {
 const {DOM} = require('react');
 const {Map} = require('immutable');
 const {isFocused} = require('./focusable');
-const {Element, Option, Field, Event} = require('./element');
+const {Element, Option, Field, Event, Attribute} = require('./element');
 const Component = require('omniscient');
 const {Deck} = require('./deck');
 const {open} = require('./web-viewer/actions');
@@ -19,6 +19,15 @@ const IFrame = Element('iframe', {
   isFocused: isFocused,
   isRemote: new Option('remote'),
   isBrowser: new Option('mozbrowser'),
+  hidden: new Field((node, current, past) => {
+    if (current != past) {
+      if (current) {
+        node.setAttribute("hidden", "true");
+      } else {
+        node.removeAttribute("hidden");
+      }
+    }
+  }),
   allowFullScreen: new Option('mozallowfullscreen'),
   src: Field((node, current, past) => {
     if (current != past) {
```

2. Change css to handle `hidden` attribute properly (meaning by it presence), which is illustrated below:

```diff
diff --git a/css/theme.css b/css/theme.css
index 80caf05..48532cb 100644
--- a/css/theme.css
+++ b/css/theme.css
@@ -36,7 +36,7 @@
   vertical-align: middle;
 }
 
-[hidden="true"] {
+[hidden] {
   display: none;
 } 
```

My preference would be to go with second option.

Gozala:
I have converted a second option to resolve issue with hidden iframes discussed above in #80. @paulrouget wanna take a look at it ? With that change this change works as expected.

paulrouget:
@Gozala r?

paulrouget:
Should I move this function into WebViewer.onOpenWindow?

Gozala:
sorry forgot to say r+

PR #79, opened 2015-02-17
Make sure git does not ignore required modules.

Gozala:
Can @paul or @gordonbrander  take a look at this.


.gitignore
```diff
@@ -1 +1,7 @@
-node_modules/
+# Lot of node developers have `node_modules` in `~/.gitignore`
+# so we use this hack to reinable it than disable everything
```
gordonbrander: Nit: it's not really a hack, but part of the way .gitignore is supposed to work. NBD, though. Merge it!

PR #80, opened 2015-02-17
Handle `hidden` attribute properly.

Gozala:
This fixes the way we style hidden attributes, see more details in #77 

@paulrouget wanna take a look at this ?

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -36,7 +36,7 @@
   vertical-align: middle;
 }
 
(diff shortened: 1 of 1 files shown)


paulrouget:
Can you explain why sometimes we see `hidden="true"` and sometimes just `hidden=""`?

paulrouget:
(apparently, you're answering that in #77)

Issue #82, opened 2015-02-18
Switching tabs should happen after releasing key


gordonbrander:
What happens:
- Right now `ctl+tab`, `cmd+shift+[`, `cmd+shift+]` all switch tabs immediately.
- There is lots of flickering jankiness

What should happen:

1. Press keyboard shortcut, tab shade is shown.
2. New active tab is highlighted
3. Page is switched when keys are released.

Hitting key command multiple times before releasing should change active highlighted thumbnail, but page should not switch until you release.

Let's make `ctl+tab` and `cmd+shift+[/]` behave the same way. Show tab shade, change active thumbnail, change active site when all keys are released.

- [x] Page should not switch until keys are released
- [x] `ctl+tab` and `cmd+shift+[` and `cmd+shift+]` should highlight next/prev tab and change location field
- [ ] Tab shade shows after short 150ms delay so you can switch between pages rapidly without seeing shade.
- [x] Hovering over tabs should change location field text

Gozala:
@gordonbrander Are you sure making `cmd+shift+[` switch the tap immediately is good idea ? To me tabswitch seems very similar to OS X app switcher & there for I have built-in expectation for it to behave the same:

`cmd+[` selects next app but only switches to it when I release `cmd`, which is pretty much what you described for `ctrl+tab`.

gordonbrander:
It's a hypothesis. The ergonomics of holding and releasing 3 keys with 2 hands is different from holding and releasing 2 keys with 1 hand. Try it in the current build. Do you think it feels ok? Bad? Why?

However, we can start with having same behavior in both instances and refine from there.

Gozala:
@gordonbrander actually dismiss what I've said `cmd+[` does nothing on OSX I was thinking of `cmd+tab`.

Gozala:
Worked under #87

gordonbrander:
- Colors should change only when page changes
- Tab shade should show after 150ms delay. So if I `ctl+tab` quickly, the page is immediately switched. If I hold down `ctl+tab` for 150ms, then tab shade is shown

Gozala:
me and @gordonbrander spend quite some time looking into the issue that was reproducible by clicking `control tab` keys really fast. We had different guess what could have being cousin it & there is an archive of the discussion here https://browserhtml.slack.com/archives/general/p1424744707000607


After looking into this further, I can say I have a logical explanation of observed behavior. This is what is going on there. To start with here a trace of the events when the bug occurs:

```js
[...]

Issue #81, opened 2015-02-18
Tab shade should only show if > tab


gordonbrander:
Hovering over the URL bar should only show tab shade if > 1 tab exists.

Gozala:
Worked under #87

Gozala:
Fixed by #87 @gordonbrander and I chatted what should be a behavior of tab key if there is only one tab,  right now it does nothing, but @gordonbrander if you want to change it I'd suggest separate issue for that.

gordonbrander:
You're right. This was a bad idea. My apologies. I'll open a new bug to change behavior.

PR #86, opened 2015-02-18
Visualise selected tab.

Gozala:

PR #87, opened 2015-02-18
Tab switch interaction tweaks #81, #82

Gozala:
- Show tabstrip if >1 tab & input is hover.
- Show tabstrip if >1 tab & ctrl+tab / ctrl+shift+tab is pressed.
- Hide tabstrip if <1 tab.
- Hide tabstrip if ctrl is released.
- Hide tabstrip if kill zone is hovered.


Gozala:
This looks pretty awkward without #86 so I'd say we need that one first.

Gozala:
@gordonbrander do you want to take a look at this ?

paulrouget:
r+

PR #93, opened 2015-02-18
Add hardcoded color themes

gordonbrander:
New commit on top of master, incorporating suggestions from
https://github.com/mozilla/browser.html/pull/67

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -18,6 +18,7 @@ const {zoomIn, zoomOut, zoomReset, open,
 const {focus, showTabStrip, hideTabStrip} = require('./actions');
 const {selectedIndex, selectNext, selectPrevious,
        remove, toggle, append, select} = require('./deck/actions');
+const {readTheme} = require('./theme');
 
 const getOwnerWindow = node => node.ownerDocument.defaultView;
 // Define custom `main` element with a custom `scrollGrab` attribute
@@ -105,11 +106,14 @@ const Browser = Component(options => {
   const isTabStripVisible = tabStrip.get('isActive') &&
                             webViewers.count() > 1;
 
+  const theme = readTheme(webViewer);
+
   return  Main({
     os: options.get('os'),
     title: webViewer.get('uri'),
     scrollGrab: true,
     className: 'moz-noscrollbars' +
+               (theme.isDark ? ' isdark' : '') +
                (options.get('isDocumentFocused') ? ' windowFocused' : '') +
                (isTabStripVisible ? ' showtabstrip' : ''),
     onDocumentFocus: event => options.set('isDocumentFocused', true),
@@ -126,6 +130,7 @@ const Browser = Component(options => {
       title: webViewer.get('title'),
     }),
     DOM.div({key: 'tabstrip',
+             style: {backgroundColor: theme.background},
              className: 'tabstripcontainer'}, [
       Tab.Deck({key: 'tabstrip',
                 className: 'tabstrip',
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -126,11 +126,11 @@ const Browser = Component(options => {
   }, [
     NavigationPanel({
       key: 'navigation',
-      input, webViewer, tabStrip,
+      input, webViewer, tabStrip, theme,
       title: webViewer.get('title'),
     }),
     DOM.div({key: 'tabstrip',
-             style: {backgroundColor: theme.background},
+             style: theme.tabstrip,
              className: 'tabstripcontainer'}, [
       Tab.Deck({key: 'tabstrip',
                 className: 'tabstrip',
--- a/src/browser/navigation-panel.js
+++ b/src/browser/navigation-panel.js
@@ -13,18 +13,24 @@ const {Element} = require('./element');
 const {navigateTo, showTabStrip, focus} = require('./actions');
 const {KeyBindings} = require('./keyboard');
 const url = require('./util/url');
+const {readTheme} = require('./theme');
 
-const WindowControls = Component(options =>
-  DOM.div({className: 'windowctrls'}, [
+const WindowControls = Component(({webViewer}) => {
+  const theme = readTheme(webViewer);
+
+  return DOM.div({className: 'windowctrls'}, [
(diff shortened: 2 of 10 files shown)


gordonbrander:
@paulrouget or @Gozala r?

gordonbrander:
The Travis build is failing because of whitespace I did not touch. Can we fix the build linter?

gordonbrander:
More discussion on these unrelated lint issues that are causing Travis failure https://github.com/mozilla/browser.html/pull/76

src/browser/navigation-panel.js
```diff
@@ -13,18 +13,24 @@ const {Element} = require('./element');
...
+const {readTheme} = require('./theme');
 
-const WindowControls = Component(options =>
-  DOM.div({className: 'windowctrls'}, [
+const WindowControls = Component(({webViewer}) => {
+  const theme = readTheme(webViewer);
+
+  return DOM.div({className: 'windowctrls'}, [
     DOM.div({className: 'windowctrl win-close-button',
+             style: {backgroundColor: theme.close},
```
Gozala: I was mainly interested in `readTheme` function so I would not have to say `style:{backgroundColor: theme.close}` but instead say a `style: theme.closeButton`, so that if in a future if we decide to style more than just `backgroundColor` we would have to only update `readTheme` function.

gordonbrander: Are you sure we want that? Some of these styles are shared between multiple components (e.g. `theme.buttons`). I can do that if you really think its the right way forward.

src/browser/navigation-panel.js
```diff
@@ -34,22 +40,26 @@ const WindowControls = Component(options =>
...
     onMouseEnter: event => tabStrip.set('isActive', true)
   }, [
     DOM.div({className: 'backbutton',
+             style: {color: theme.buttons},
              key: 'back',
              onClick: event => webViewer.set('readyState', 'goBack')}),
     InputField({
       key: 'input',
       className: 'urlinput',
+      style: {color: theme.domain},
```
Gozala: `style: theme.input`

src/browser/navigation-panel.js
```diff
@@ -68,9 +78,12 @@ const NavigationControls = Component(({input, tabStrip, webViewer, title}) =>
       DOM.span({key: 'identity',
                 className: 'identity'}, ''),
       DOM.span({key: 'location',
+                style: {color: theme.domain},
```
Gozala: `style: theme.address` or `style: theme.location`

src/browser/navigation-panel.js
```diff
@@ -78,15 +91,19 @@ const NavigationControls = Component(({input, tabStrip, webViewer, title}) =>
     ]),
     DOM.div({key: 'reload-button',
              className: 'reloadbutton',
+             style: {color: theme.buttons},
```
Gozala: I'd do `style: theme.reloadButton` instead of `style: theme.button` so it's up to `readTheme` to decide if all buttons share same style or some of them are special.

src/browser/navigation-panel.js
```diff
@@ -94,7 +111,7 @@ const NavigationPanel = Component(({key, input, tabStrip, webViewer, title}) =>
                         + (webViewer.get('securityState') == 'secure' ? ' ssl' : '')
                         + (webViewer.get('securityExtendedValidation') ? ' sslv' : '')
   }, [
-    WindowControls({key: 'controls'}),
+    WindowControls({key: 'controls', webViewer}),
```
Gozala: I think it would be better to do `location: webViewer.get('location')` instead to avoid generating `WindowControls` even if nothing about the location changes.

gordonbrander: We need webViewer to generate theme object.

src/browser/web-viewer.js
```diff
@@ -175,14 +175,13 @@ WebViewer.onLoadStart = state => event => state.merge({
...
-  backgroundColor: event.detail.backgroundColor
+  isLoading: false
 });
 
 WebViewer.onTitleChange = state => event => state.set('title', event.detail);
 
-WebViewer.onLocationChange = state => event =>
-  state.set('location', event.detail);
+WebViewer.onLocationChange = state => event => state.merge(
+  Object.assign({location: event.detail}, getHardcodedColors(event.detail)));
```
Gozala: `{location: event.detail, theme: readTheme(event.detail)}`

src/browser/browser.js
```diff
@@ -105,11 +106,14 @@ const Browser = Component(options => {
   const isTabStripVisible = tabStrip.get('isActive') &&
                             webViewers.count() > 1;
 
+  const theme = readTheme(webViewer);
```
Gozala:
```js
const theme = readTheme(webViewer.get('colors'));
```

src/browser/browser.js
```diff
@@ -126,6 +130,7 @@ const Browser = Component(options => {
       title: webViewer.get('title'),
```
Gozala: theme

src/browser/browser.js
```diff
@@ -126,6 +130,7 @@ const Browser = Component(options => {
...
     }),
     DOM.div({key: 'tabstrip',
+             style: {backgroundColor: theme.background},
```
Gozala:
```js
style: theme.tabstrip
```

src/browser/navigation-panel.js
```diff
@@ -34,22 +40,26 @@ const WindowControls = Component(options =>
...
-  ]));
+  ])});
 
 const inputBindings = KeyBindings({'escape': focus});
 
 
-const NavigationControls = Component(({input, tabStrip, webViewer, title}) =>
-  DOM.div({
+const NavigationControls = Component(({input, tabStrip, webViewer, title}) => {
+  const theme = readTheme(webViewer);
```
Gozala:
```js
const theme = readTheme(webViewer.get('location') || webViewer.get('uri'))
```

src/browser/navigation-panel.js
Gozala: WindowControls({key: 'controls', theme}),

Issue #72, opened 2015-02-16
Figure out the right progress bar behavior


paulrouget:
Bug 1125713 will expose progress events to the browser API. We get the `current` and `total` values. Both of these values change over time. We get some weird results sometimes where `current > total`.

Some examples:

```
github.com:

"9699ms: indeterminate"
"9723ms: 120% 56554 / 46969"
"9734ms: 109% 111157 / 101572"
"9746ms: 109% 113467 / 103882"
"9754ms: 109% 114983 / 105398"
"9765ms: 104% 221961 / 212376"
"9772ms: 101% 700145 / 690560"
"9861ms: 101% 700413 / 690828"
"9868ms: 101% 700666 / 691081"
"9874ms: 101% 703358 / 693773"
"9944ms: 101% 704546 / 693773"
"9961ms: 101% 722038 / 711265"
"9981ms: 101% 832803 / 822030"
"10057ms: 101% 935609 / 924836"
"10223ms: 101% 938501 / 927728"
"10230ms: 101% 938536 / 927763"
"10491ms: 101% 938571 / 927798"

bbc.com:

"12046ms: 23% 33146 / 138241"
"12318ms: 24% 34503 / 138241"
"12327ms: 30% 42645 / 138241"
"12339ms: 33% 48361 / 143957"
"12363ms: 34% 49718 / 143957"
"12375ms: 39% 56503 / 143957"
"12391ms: 41% 59217 / 143957"
"12948ms: 66% 95856 / 143957"
"12965ms: 70% 101284 / 143957"
"12979ms: 70% 102304 / 144977"
"12985ms: 70% 102492 / 145165"
"12990ms: 73% 119869 / 162542"
"12995ms: 74% 125179 / 167852"
"13000ms: 75% 130319 / 172992"
"13006ms: 76% 135139 / 177812"
"13011ms: 80% 174366 / 217039"
"13017ms: 81% 185314 / 227987"
"13023ms: 81% 190767 / 233440"
"13028ms: 82% 206575 / 249248"
"13034ms: 83% 221369 / 264042"
"13040ms: 85% 246018 / 288691"
"13045ms: 85% 249935 / 292608"
"13051ms: 85% 253549 / 296222"
"13056ms: 85% 256366 / 299039"
"13072ms: 85% 259466 / 302139"
"13086ms: 86% 263972 / 306645"
"13094ms: 86% 269261 / 311934"
"13103ms: 86% 272525 / 315198"
[...]

PR #102, opened 2015-02-19
better progressbar

paulrouget:
issue #72. Please only look at the last commit (`wip`).

This is still WIP. But I need to know if this is the right approach.

Here is how it works:

The progressbar is split in 3 zones: **`[ aaa bbbb ccccccc]`**.

Zone **A** (~10% of the locationbar), is blue as soon as we start loading a page.
Zone **B** (~40% of the locationbar), is slowly incremented as the page is loading (see explanation below).
Zone **C** (~50%, rest of the locationbar), is filled once the page is fully loaded.

Zone **B** is updated every time gecko tells us there's a progression. But we *do not* use at the values of the progress event (see #72). So the bar moves only if there's activity, and its position is defined by a invert tangent function of loading time, which means as time pass, less and less pixels are added to the bar, which make it impossible to actually fill the all 40% reserved for this zone.

Patch is not finished. There are issues with tab switching. But before finishing it, I'd appreciate some feedback.

@gordonbrander ?

### DIFF
--- a/src/browser/progressbar.js
+++ b/src/browser/progressbar.js
@@ -1,0 +1,112 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Progress bar logic:
+ * The progressbar is split in 3 zones. Called A, B and C.
+ *   Zone A is slowly filled while the browser is connecting to the server.
+ *   Zone B is slowly filled while the page is being downloaded.
+ *   Zone C is filled once the page has loaded (fast).
+ * Zone A and B get filled slower and slower in a way that they are never 100% filled.
+ */
+
+define((require, exports, module) => {
+
+  'use strict';
+
+  const {Element, Field} = require('./element');
+  const Component = require('omniscient');
+
+  const ProgressBarElement = Element('div', {
+    progressBarColor: Field((node, current, past) => {
+      if (current != past) {
+        node.style.setProperty('--progressbar-color', current);
+      }
+    }),
+  });
+
+  const PI = 3.1416; // > PI
+
+  // Animation parameters:
+  const A = 0.2;              // Zone A size (a full progress is equal to '1'
+  const B = 0.3;              // Zone B size
+  const APivot = 200;         // When to reach ~80% of zone A
+  const BPivot = 500;         // When to reach ~80% of zone B
+  const CSpeed = 0.002;       // Speed to fill zone C (how fast we reach '1')
+  const StartFading = 0.8;    // When does opacity starts decreasing to 0
+
+  // Inverse tangent function: [0 - inf] -> [0 - PI/2]
+  // ApproachFunc: [time, pivot] -> [0 - 1]
+  // Pivot value is more or less when the animation seriously starts to slow down
+  const ApproachFunc = (tMs, pivoMs) => 2 * Math.atan(tMs / pivoMs) / PI;
+
+  // rfa and before are global and not associated to any state. They are used
+  // to drive the progressbar animation. There's only one animation to play
+  // at the same time.
+  let rfa = -1;
+  let before = 0;
+
+  const ProgressBar = Component([{
+    step(now) {
+      const viewer = this.props.webViewerCursor;
+
+      let progress; // value between 0 and 1
+
+      if (viewer.get('isLoading')) {                                            // Zone A
+        const startLoadingTime = viewer.get('startLoadingTime');
+        if (viewer.get('isConnecting')) {
+          progress = A * ApproachFunc(now - startLoadingTime, APivot);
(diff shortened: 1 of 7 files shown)


gordonbrander:
@paulrouget the behavior you describe sounds great.

If possible, let's try to do this in such a way that we can tune the width of the A, B and C zones and the speed of the progress bar in those zones. I would especially like to get the feel of speed in the C zone. 

CogSci: people remember the _last_ part of an experience most. A pleasant surprise at the end of waiting will make you forget you ever waited.

paulrouget:
Not ready for review. But I need feedback.

Took much longer than expected. I tried too hard to use a pure CSS approach. I haven't found any correct way to achieve the animations we wanted. Mostly because we use one single progressbar for all the iframes. It required a lot of CSS voodoo to get it right, and then there was the issue with tab switching and remembering exactly the state of the progress.

@Gozala Can you please look at how I use Component and Element. I'm not sure it's the best way to do it.

@gordonbrander Can you please try this patch, and also look at the progressbar algorithm and tell me if this works for you (see the parameters we can tweak).

paulrouget:
Also, how do we do for the theme?

gordonbrander:
@paulrouget looking at the update. It's working, but oddly enough, the visual result is indistinguishable from a linear 5s animation. I'll play with the settings and see what I get.

gordonbrander:
Ah, it's because I'm still seeing the CSS animated 5s linear progress bar gradient. D'oh!  I hacked on the code for a while, but couldn't get the progress bar to show up.

@paulrouget the animation constants look good to me. Without eyeballing it, I can't confirm that it's perfect, but the direction looks good to me.

src/browser/progressbar.js
```diff
@@ -0,0 +1,106 @@
...
+  const CSpeed = 0.002;       // Speed to fill zone C (how fast we reach '1')
+  const StartFading = 0.8;    // When does opacity starts decreasing to 0
+
+  // Inverse tangent function: [0 - inf] -> [0 - PI/2]
+  // ApproachFunc: [time, pivot] -> [0 - 1]
+  // Pivot value is more or less when the animation seriously starts to slow down
+  const ApproachFunc = (tMs, pivoMs) => 2 * Math.atan(tMs / pivoMs) / PI;
+
+  // requestAnimationFrame ID
+  let rfa = -1;
```
Gozala: Only thing that I don't like here is this `rfa` global, which implies we can render only a single ProgressBar at a time. Also makes me wonder what the behavior is gonna be like when multiple tabs are going to be loading concurrently and updating their progress.

src/browser/progressbar.js
```diff
@@ -0,0 +1,106 @@
...
+      if (progress < 1) {
+        rfa = requestAnimationFrame(() => this.step());
+      } else {
+        rfa = -1;
+      }
+      this.props.selectedWebViewerCursor = viewer.merge({ progress, progressTime: now });
+    },
+    componentDidUpdate() {
+      if (this.props.selectedWebViewerCursor.get('progress') < 1) {
+        if (rfa < 0) {
```
Gozala: Is this to avoid scheduling rfa if it already is scheduled ?

src/browser/web-viewer.js
```diff
@@ -63,6 +64,9 @@ define((require, exports, module) => {
   WebViewer.onLoadStart = webViewerCursor => event => webViewerCursor.merge({
     readyState: 'loading',
     isLoading: true,
+    connecting: true,
```
Gozala: Can you call this `isConnecting` to follow the used name convention ?

src/browser/web-viewer.js
```diff
@@ -63,6 +64,9 @@ define((require, exports, module) => {
...
+    startLoadingTime: performance.now(),
```
Gozala: Is `event.timeStamp` not precise enough ?

paulrouget:
Sorry… I managed to miss a commit.

src/browser/progressbar.js
paulrouget: We want only one single RFA. We don't want to track the progress of hidden iframes. This loop needs to be only active for the active tab.

paulrouget: I'll add a comment or try to make it clearer what's going on here.

src/browser/web-viewer.js
paulrouget: We need to use one consistent way to measure time. `timeStamp` is available from events. We can't use this value with `performance.now()` (from RFA).

src/browser/progressbar.js
paulrouget: Wait. You're absolutely right. There's something wrong here. RFA is never canceled.

paulrouget: I was confused :) RFA doesn't need to be canceled.

paulrouget:
This might not be perfect. I noticed some glitches (I believe it's gfx bugs), and maybe the global `rfa` is not satisfying, but let's move forward and file issue to improve this mechanism. Always better than what we had before.

Issue #98, opened 2015-02-19
Move thumbnails into mozilla/browser.html repo


paulrouget:
We should move thumbnails from https://github.com/gordonbrander/website-tiles to this repo.

gordonbrander:
Closed by #103

PR #103, opened 2015-02-19
add tiles

paulrouget:
See #98

I don't really want to use a submodule.

@gordonbrander works for you?


gordonbrander:
@paulrouget fine by me as long as you're ok with adding more of these to our repo in future.

paulrouget:
Of course.

PR #104, opened 2015-02-19
Fix ssl character alignment

paulrouget:
Switching between ssl and non-ssl page makes the navbar jumbs a bit.

@gordonbrander, can you review this?


gordonbrander:
This looks better.

Note: the lock is about 1px below where it should be, but I think that was the case beforehand. Looks like an issue with the baseline of the icon font to me. We can deal with it later.

Issue #107, opened 2015-02-19
Restore tabs after restart


gordonbrander:
What it says.

gordonbrander:
Closed by https://github.com/mozilla/browser.html/commit/fd8add2637272a362cd5b2f693317c88e3123f32

PR #116, opened 2015-02-19
Implement session restore

Gozala:
Fix for the #107 @paulrouget @gordonbrander can one of you take a look ?

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -7,6 +7,12 @@ define((require, exports, module) => {
   'use strict';
 
   const url = require('./util/url');
+  const {fromJS} = require('immutable');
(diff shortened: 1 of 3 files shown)


gordonbrander:
Looking

src/browser/actions.js
```diff
@@ -7,6 +7,12 @@ define((require, exports, module) => {
   'use strict';
 
   const url = require('./util/url');
+  const {fromJS} = require('immutable');
+  const {open} = require('./web-viewer/actions');
+  // TODO: Shuld be `const {version} = require('package.json`);` instead but require.js
```
gordonbrander: Nit: typo. `Should be...`

PR #127, opened 2015-02-20
Work in progress tab switch on release

Gozala:
This pull is **WORK IN PROGRESS** implementation of tab switch preview described in #82 that updates address bar and active tile but only switches tab on key release.

Publishing it earlier so that @gordonbrander can play around with it to see how it feels. I don't find this behavior more pleasant.

### DIFF
--- a/css/tabstrip.css
+++ b/css/tabstrip.css
@@ -67,10 +67,12 @@ main.showtabstrip iframes {
   margin-right: 0;
 }
 
+/*
 .tab:hover {
   transform: scale(1);
 }
+*/
 
-.tabstrip:not(:hover) .tab.selected {
+.tab.selected {
   transform: scale(1);
 }
--- a/main.js
+++ b/main.js
@@ -5,6 +5,7 @@ require.config({
   paths: {
     browser: '../src/browser',
     shims: '../src/shims',
+    lang: '../src/lang',
     // http://facebook.github.io/react/
     react: 'react/dist/react',
     // http://facebook.github.io/immutable-js/
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -8,7 +8,7 @@ define((require, exports, module) => {
 
   const Component = require('omniscient');
   const {DOM} = require('react');
-  const {compose} = require('lang/functional');
+  const {compose, throttle} = require('lang/functional');
   const {NavigationPanel} = require('./navigation-panel');
   const {WebViewer} = require('./web-viewer');
   const {Tab} = require('./page-switch');
@@ -17,8 +17,8 @@ define((require, exports, module) => {
   const {zoomIn, zoomOut, zoomReset, open,
          goBack, goForward, reload, stop, title} = require('./web-viewer/actions');
   const {focus, showTabStrip, hideTabStrip,
-         writeSession, resetSession} = require('./actions');
-  const {indexOfSelected, indexOfActive,
+         writeSession, resetSession, resetSelected} = require('./actions');
+  const {indexOfSelected, indexOfActive, isActive,
          selectNext, selectPrevious, select, activate,
          previewed, remove, append} = require('./deck/actions');
   const {readTheme} = require('./theme');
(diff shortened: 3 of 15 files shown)


paulrouget:
I think its the default content style:

```
background: white linear-gradient(rgba(115,115,115,0.5) 0, rgba(215,215,215,0.5) 3px, rgba(255,255,255,0.2) 16px);
```
We should use FIrefox' content.css, not the one provided by b2g.

gordonbrander:
Yeah, I strongly feel this is the right approach. However, we should not change the color of the UI until the page is switched (keys are released).

css/tabstrip.css
```diff
@@ -67,10 +67,12 @@ main.showtabstrip iframes {
   margin-right: 0;
 }
 
+/*
```
gordonbrander: Should this be commented out? Remove or un-comment.

main.js
```diff
@@ -5,6 +5,7 @@ require.config({
   paths: {
     browser: '../src/browser',
     shims: '../src/shims',
+    lang: '../src/lang',
```
gordonbrander: Nit: I mistook lang to mean `i10n`

src/browser/browser.js
```diff
@@ -67,27 +72,46 @@ define((require, exports, module) => {
...
+    items.count() > 1 ? remove(items, p) :
+    items.set(0, open({isSelected: true, isActive: true}));
+
+  const closeTab = item =>
+    close(x => x.get('id') == item.get('id'));
+
 
   const edit = edit => cursor => cursor.update(edit);
+
+  const onSelectNext = throttle(edit(selectNext), 200)
```
gordonbrander: Just checking: I thought we had to remove `throttle` because it was causing state bug due to delay. Did you manage to fix issue in a way that makes it safe to include `throttle`?

src/browser/core.js
```diff
@@ -17,6 +17,32 @@ define(function(require, exports, module) {
   var render = (Component, initial, target) => {
     let current = null;
 
+    const schedule = (task) => {
+      if (!schedule.d) {
+        schedule.d = true;
+
+        // Disable anymation framing as it couses some
+        // misbehavior (see https://github.com/mozilla/browser.html/issues/82)
+        //window.requestAnimationFrame(() => {
```
gordonbrander: Do we mean to remove requestAnimationFrame? Should we disable it only sometimes?
We should remove this code or un-comment it.

src/browser/core.js
```diff
@@ -45,15 +71,16 @@ define(function(require, exports, module) {
...
+        console.log('! update', path.join('.'), current.toJSON());
+      }
+
+
+      schedule(draw);
     }
 
     // Spawn render loop by stepping into!
-    step(initial, null);
+    step(initial, null, []);
```
gordonbrander: Should empty array be here?

src/browser/deck/actions.js
```diff
@@ -6,119 +6,218 @@ define((require, exports, module) => {
...
-           items.slice(0, before - 1).push(item).concat(items.slice(before - 1));
+  const activateNext = compose(activate, selectNext);
+  const activatePrevious = compose(activate, selectPrevious);
+
+  // Take an `items` and optional `p` predicate function and removes
+  // first item where `p(item)` is true. If item to be removed happens
+  // to be selected or active then it takes care of activating / selecting
+  // different item according to logic explained in the inline comments
+  // below.
+  const remove = (items, p=exports.isActive) => {
```
gordonbrander: Nit: since `isActive` is a const in scope, it would be safer to reference `isActive` directly.

src/browser/web-viewer.js
```diff
@@ -56,14 +56,16 @@ define((require, exports, module) => {
...
 
-  WebViewer.onFocus = webViewerCursor => event => webViewerCursor.set('isFocused', true);
+  WebViewer.onFocus = webViewerCursor => event =>
+    webViewerCursor.set('isFocused', true);
 
   WebViewer.onLoadStart = webViewerCursor => event => webViewerCursor.merge({
     readyState: 'loading',
     isLoading: true,
-    icons: null,
+    icons: {},
```
gordonbrander: What is the difference?

src/browser/browser.js
Gozala: > Just checking: I thought we had to remove throttle because it was causing state bug due to delay. Did you manage to fix issue in a way that makes it safe to include throttle?

I do believe issue was actually caused by an animation frame issue that I've explained in #82 if that turns out not to be a case we can remove throttle as well.

src/browser/core.js
Gozala: > Should empty array be here?

It's so that that code that does logging would always have a path to print, without it's going to throw.

src/browser/deck/actions.js
Gozala: > Nit: since isActive is a const in scope, it would be safer to reference isActive directly.

That's what I had originally but it turns out that `isActive` declaration in the function body makes such a function illegal, does not makes sense if you ask me, but...

PR #129, opened 2015-02-20
Ignore .DS_Store files

gordonbrander:
Add .DS_Store to .gitignore. .DS_Store is one of those junk OSX files. I'm tired of seeing them show up in
the commit staging list.


paulrouget:
(for these kind of changes, feel free to land it without a review)

PR #130, opened 2015-02-20
Fix interactable gap between thumbnails

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/64

Also increases interactable area of thumbnails.

Note that I first tried using only CSS (`::after` and `attr()`) to solve this problem, but our support of CSS generated is too limited to pull this off. So instead, I added an element.

### DIFF
--- a/css/tabstrip.css
+++ b/css/tabstrip.css
@@ -48,15 +48,22 @@ main.showtabstrip iframes {
 }
 
 .tab {
-  margin: 15px 0 0;
-  height: 62.5px;
+  /* Keep buffer space around the visual thumbnail to make them easier
+  to click. */
+  height: 100px;
   width: 100px;
+}
+
+.tab-thumb {
+  background: white;
+  background-size: 100% auto;
   border: 1px solid rgba(0,0,0,0.07);
   border-radius: 4px;
-  background-color: white;
+  height: 62.5px;
+  margin: 16px 0 0;
   transform: scale(0.8);
   transition: transform 150ms ease;
-  background-size: 100% 100%;
+  width: 100px;
 }
 
 .tab:first-of-type {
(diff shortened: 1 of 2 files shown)


gordonbrander:
@Gozala r?

src/browser/page-switch.js
```diff
@@ -32,7 +29,13 @@ define((require, exports, moudle) => {
           remove(items, equals(item));
         }
       }
-    }));
+    }, [
+      DOM.span({
+        key: `thumb-${item.get('id')}`,
```
Gozala: Nit: You don't need to include id in the key, keys just need to be unique with-in it's siblings not across the whole tree.

css/tabstrip.css
```diff
@@ -67,10 +74,10 @@ main.showtabstrip iframes {
   margin-right: 0;
 }
 
-.tab:hover {
+.tab:hover > .tab-thumb {
```
Gozala: Nit: I would prefer full names, like `.tab-thumbnail` over shorthands, also I think direct descendent selectors `.tab:hover > .tab-thumbnail` are slower than non direct `.tab:hover .tab-thumbnail`.

Gozala:
Looks good to me, leaving up to you either merge or address nits & then merge.

css/tabstrip.css
gordonbrander: Direct descendent are actually supposed to be faster https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Writing_efficient_CSS. Though if we JIT CSS like Chrome, it shouldn't really matter.

At some point, I would like to rejigger the CSS so we use naming conventions instead of chaining selectors.

gordonbrander: I'm removing them, though. Premature optimization.

PR #134, opened 2015-02-22
Animate background color for tabstrip and navbar.

### LINKED ISSUE #132: Animate background color changes
Animate background color changes for themes.
This is probably just a CSS change.

KevinGrandon:
Fixes #132 

@gordonbrander - It looks like there was already a 200ms transition elsewhere in the codebase, so I went with the same here. Thoughts?

I'm also having trouble with cache invalidation and using the HTTPServer here. Do you guys have any suggestions for that?

### DIFF
--- a/css/navbar.css
+++ b/css/navbar.css
@@ -4,6 +4,7 @@
   padding: 5px;
   justify-content: space-between;
   scroll-snap-coordinate: 0 0;
+  transition: background-color 200ms ease;
 }
 
 .freeendspacer {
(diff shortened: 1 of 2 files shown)


paulrouget:
> I'm also having trouble with cache invalidation and using the HTTPServer here. Do you guys have any suggestions for that?

When you do F5 cache is not always cleared, and restarting doesn't solve the problem? I have the same issue, I haven't investigated yet.

KevinGrandon:
> When you do F5 cache is not always cleared, and restarting doesn't solve the problem? I have the same issue, I haven't investigated yet.

Yeah, I can't seem to figure out how to blow away the cache. It seems to persist across a restart of the server and b2g instance as well.

paulrouget:
Yep. No good solution other than deleting the profile.

PR #136, opened 2015-02-24
Always show tab strip on location hover

gordonbrander:
Even if there is only one website. Closes
https://github.com/mozilla/browser.html/issues/131.


gordonbrander:
@Gozala r? This is a small change and seems to work great, but I want to make sure this state logic doesn't appear in more than one place.

Gozala:
I think that's only place, so it should be good.

Issue #108, opened 2015-02-19
Scroll should un-focus URL bar


gordonbrander:
Or alternative: focus could instead keep URL bar visible. @Gozala whichever you think is best.

Gozala:
@paulrouget I wanted to tackle this one, but I'm afraid I was unable to figure out how our address bar is even happens to appear / disappear. I would like to add something like `navigationPanel: {isActive: false}` to a global state where `isActive` is true if bar is visible and `false` if hidden.

Gozala:
@paulrouget ok I think I found what I was looking for, also have submitted #143 to fix it.

paulrouget:
Yeah, you need to do that on scroll. But that won't work well once we will switch to apzc. Well… maybe it would. I don't exactly know how it works: does a scroll event handler will make apzc bail or just call the listener later?

(I asked the question, I'll edit later once I know)

> Scroll should un-focus URL bar or focus could instead keep URL bar visible

@gordonbrander we need to know what's best, because it's really 2 different strategies.

gordonbrander:
I think "focus should instead keep URL bar visible" is better.

But, __I think we should defer this item__.
I'm looking at it in the current version of master and with other interaction bugs fixed, it's a very minor issue at best. I don't feel it will impact the MVP.

paulrouget:
(We can use `onscroll`)

Gozala:
so here is a second implementation that keeps url bar instead. I personally find loose focus on scroll behavior a lot more.

gordonbrander:
Closed by https://github.com/mozilla/browser.html/commit/088648802150e5c17493ac90ede40ac86292c71b

PR #143, opened 2015-02-25
Focus content on scroll out #108

Gozala:
Fix for #108

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -132,6 +132,13 @@ define((require, exports, module) => {
 
     const theme = readTheme(activeWebViewerCursor);
 
+    const onNavigationPanelVisible = () => {
+    }
+
+    const onNavigationPanelHidden = () => {
+      focus(activeWebViewerCursor);
+    }
+
     return Main({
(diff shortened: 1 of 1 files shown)


src/browser/browser.js
```diff
@@ -132,6 +132,13 @@ define((require, exports, module) => {
 
     const theme = readTheme(activeWebViewerCursor);
 
+    const onNavigationPanelVisible = () => {
```
gordonbrander: Noop?

paulrouget:
I'm not sure this will work well with apzc.

Gozala:
> I'm not sure this will work well with apzc.

@paulrouget Can we deal with that if it / when it won't ? Also I could not think of any better way dinging it, if you have a suggestion I'm all ears.

paulrouget:
> @paulrouget Can we deal with that if it / when it won't ? Also I could not think of any better way dinging it, if you have a suggestion I'm all ears.

It will work with APZC. But see Gordon comment in #108.

Gozala:
We wend with #147 instead

PR #157, opened 2015-02-27
Move progress bar out of location field

gordonbrander:
I'm working on another visual solution, since this one is pretty generic, but this is a good interim fix for https://github.com/mozilla/browser.html/issues/152 and https://github.com/mozilla/browser.html/issues/115.

### DIFF
--- a/css/locationbar.css
+++ b/css/locationbar.css
@@ -6,7 +6,6 @@
   border-radius: 3px;
   line-height: 30px;
   background-color: rgba(0,0,0,0.07);
-  position: relative;
   overflow: hidden;
 }
 
@@ -17,12 +16,11 @@
 .progressbar {
   z-index: 99;
   display: block;
-  width: 500px;
-  height: 30px;
-  background-image: linear-gradient(110deg, var(--progressbar-color), var(--progressbar-color) 460px, transparent 460px);
-  background-size: 500px 30px;
+  width: 100%;
+  height: 3px;
+  margin-left: -100%;
   position: absolute;
-  top: 0;
+  top: 40px;
   left: 0;
 }
 
--- a/css/theme.css
+++ b/css/theme.css
@@ -45,7 +45,7 @@
 main {
   height: 100vh;
   width: 100vw;
-  overflow-y: scroll;
+  overflow-y: hidden;
   flex-flow: column;
   background-color: var(--bg);
   color: black;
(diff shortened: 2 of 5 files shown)


gordonbrander:
@Gozala r?

css/locationbar.css
```diff
@@ -17,12 +16,11 @@
...
-  width: 500px;
-  height: 30px;
-  background-image: linear-gradient(110deg, var(--progressbar-color), var(--progressbar-color) 460px, transparent 460px);
-  background-size: 500px 30px;
+  width: 100%;
+  height: 3px;
+  margin-left: -100%;
   position: absolute;
-  top: 0;
+  top: 40px;
```
paulrouget: You're covering the content? It will cover any progressbar implemented that way (like youtube.com or medium.com)

css/locationbar.css
```diff
@@ -17,12 +16,11 @@
...
   z-index: 99;
   display: block;
-  width: 500px;
-  height: 30px;
-  background-image: linear-gradient(110deg, var(--progressbar-color), var(--progressbar-color) 460px, transparent 460px);
-  background-size: 500px 30px;
+  width: 100%;
+  height: 3px;
+  margin-left: -100%;
   position: absolute;
```
paulrouget: Does it have to be `position:absolute`? Can't we keep it in the flow?

paulrouget: Does it have to be position:absolute? Can't we keep it in the flow?

css/theme.css
```diff
@@ -59,6 +59,10 @@ main * {
   display: flex;
 }
 
+main.scrollable {
+  overflow-y: scroll;
+}
+
```
paulrouget: Don't we have a PR just for this change?

css/locationbar.css
gordonbrander: Not a problem in practice because progress bar only exists when in-page progress bars do not. In other words, medium implements its own progress bar in order to _avoid_ triggering progress bar in the browser.

css/locationbar.css
gordonbrander: The intent is to overflow the content. If this is a problem in practice, we could change the design.

css/theme.css
gordonbrander: Not sure how that happened? A rebase? Will fix.

gordonbrander:
Rebased https://github.com/mozilla/browser.html/pull/161

PR #161, opened 2015-02-27
Move progress bar below tabstrip

gordonbrander:
This approach is pretty generic, but solves a lot of the color matching
issues. Could use as an interim fix. Fixes
https://github.com/mozilla/browser.html/issues/115 and
https://github.com/mozilla/browser.html/issues/152


Gozala:
I assume @gordonbrander replies satisfies @paulrouget so I'll just merge this! @paulrouget fee free to followup with @gordonbrander if you disagree.

PR #162, opened 2015-02-27
Use classSet to concat class names.

Gozala:


src/browser/navigation-panel.js
```diff
@@ -100,12 +101,15 @@ define((require, exports, module) => {
...
-                            (webViewerCursor.get('isLoading') ? ' loading' : '') +
-                            (webViewerCursor.get('securityState') == 'secure' ? ' ssl' : '') +
-                            (webViewerCursor.get('securityExtendedValidation') ? ' sslv' : '')
+      className: ClassSet({
+        navbar: true,
+        urledit: inputCursor.get('isFocused'),
+        cangoback: webViewerCursor.get('canGoBack'),
+        canreload: webViewerCursor.get('location'),
+        loading: webViewerCursor.get('isLoading'),
+        ssl: webViewerCursor.get('securityState'),
```
gordonbrander: We used to compare `webViewerCursor.get('securityState') == `secure`. What is the contents of `webViewerCursor.get('securityState')`? `secure` or `null`? Want to make sure we're not introducing a state bug.

Gozala: Good catch, I'll fix that

gordonbrander:
Looks good. One question I have is how often we're creating these class description objects? Once per render? I'm not too worried about the garbage -- after all, garbage collector is there to pick up garbage -- but it's nice to understand what is generating garbage and how much.

Gozala:
> Looks good. One question I have is how often we're creating these class description objects? Once per render? I'm not too worried about the garbage -- after all, garbage collector is there to pick up garbage -- but it's nice to understand what is generating garbage and how often.

Yes it's per render, which should be triggered only if state we're rendering has changed.

PR #169, opened 2015-03-02
Move progress bar out of location bar. Fixes #167

### LINKED ISSUE #167: Hovering the bottom edge of the title bar causes tab strip to show
Hovering anywhere to the right of the location bar causes tab strip to show. This should not be.

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/167.

### DIFF
--- a/css/locationbar.css
+++ b/css/locationbar.css
@@ -1,9 +1,11 @@
 .locationbar {
   -moz-window-dragging: no-drag;
   align-items: center;
+  align-self: center;
   flex-basis: 460px;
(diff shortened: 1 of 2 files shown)


css/locationbar.css
```diff
@@ -1,9 +1,11 @@
 .locationbar {
   -moz-window-dragging: no-drag;
   align-items: center;
+  align-self: center;
```
gordonbrander: Wait a sec. Not necessary. Pushing a fix that removes this.

PR #171, opened 2015-03-02
Support Alt-1 ... Alt-9 keyboard shortcuts.

fabricedesre:


src/browser/deck/actions.js
```diff
@@ -196,6 +196,22 @@ define((require, exports, module) => {
...
   const prepend = (items, item) => insert(items, item, 0);
 
+  const maybeGoToTab = (items, index) => {
+    if (index > items.length) {
+      return items;
+    }
+
+    const selectTab = advance(switchSelected,
+                              indexOfSelected,
+                              () => { return index - 1; });
```
paulrouget: ` _ => index - 1`

paulrouget:
On Mac, we need to use `meta`, not tab.
Can you rename `goToLastTab` to `selectedLat`, `maybeGoToTab` to `maybeSelectIndex` and not use start at `0`, not `1` (drop `index - 1`).

fabricedesre:
I fixed the naming and off by one thing.

For the meta/alt, it's the same issue we have with alt + left to go back in history, ie. we need a better way to express os specific bindings. Can we do that in a follow-up?

src/browser/browser.js
```diff
@@ -67,7 +68,7 @@ define((require, exports, module) => {
     'accel right': goForward,
     'escape': stop,
     'accel r': reload,
-    'F5': reload
+    'F5': reload,
```
gordonbrander: Accidental comma?

src/browser/browser.js
```diff
@@ -101,7 +104,16 @@ define((require, exports, module) => {
...
+    'ctrl pageup': onSelectPrevious,
+    'alt 1': edit(goToTab(0)),
+    'alt 2': edit(goToTab(1)),
+    'alt 3': edit(goToTab(2)),
+    'alt 4': edit(goToTab(3)),
+    'alt 5': edit(goToTab(4)),
+    'alt 6': edit(goToTab(5)),
+    'alt 7': edit(goToTab(6)),
+    'alt 8': edit(goToTab(7)),
+    'alt 9': edit(selectLast),
```
gordonbrander: Same

src/browser/deck/actions.js
```diff
@@ -196,6 +196,22 @@ define((require, exports, module) => {
...
+    }
+
+    const selectTab = advance(switchSelected,
+                              indexOfSelected,
+                              () => index);
+    // First select the new tab, then activate it.
+    return compose(activate, selectTab)(items);
+  }
+
+  const selectLast = (items) => {
```
gordonbrander: Nit: could be short form.

src/browser/browser.js
fabricedesre: that's not an array, and that makes cleaner diffs when you add items.

src/browser/deck/actions.js
```diff
@@ -196,6 +196,20 @@ define((require, exports, module) => {
 
   const prepend = (items, item) => insert(items, item, 0);
 
+  const maybeSelectIndex = (items, index) => {
+    if (index >= items.length) {
+      return items;
+    }
+
+    const selectTab = advance(switchSelected,
```
Gozala: You could just do: `return activate(switchSelected(items, indexOfSelected(items), index));`

src/browser/deck/actions.js
```diff
@@ -196,6 +196,20 @@ define((require, exports, module) => {
 
   const prepend = (items, item) => insert(items, item, 0);
 
+  const maybeSelectIndex = (items, index) => {
```
Gozala: Since you do activate a tab, please do not call this, `*select*`, but `*activate*` instead.

src/browser/deck/actions.js
```diff
@@ -196,6 +196,20 @@ define((require, exports, module) => {
...
+    }
+
+    const selectTab = advance(switchSelected,
+                              indexOfSelected,
+                              () => index);
+    // First select the new tab, then activate it.
+    return compose(activate, selectTab)(items);
+  }
+
+  const selectLast = items => maybeSelectIndex(items, items.length - 1);
```
Gozala: Here ase well please call `activateLast` instead.

src/browser/deck/actions.js
Gozala: Just thinking it out loud: It maybe better idea to just update `activate` to:

```js
activate = (items, item=selected(items)) => ....
```

That way `activate` API signature will match `select` API and there will be no need for additional `activateLast` as `goToTab` (btw I'd call it switchToTab`) can be:

```js
const goToTab = index => items => activate(items, items.get(index));
```

Note that if index is out of bound it will return undefined and there for activate selected tab, which is going to be noop.

As of  `alt 9` it will become:

```js
activate(items, items.last());
```

PR #172, opened 2015-03-03
Do not store progress at the viewer level

paulrouget:
@Gozala , I was thinking, can't we just pass the whole state to each component? Every time we add a new parameter, it makes lines larger, and we have to add the parameter to all the parents.

### DIFF
--- a/src/browser/progressbar.js
+++ b/src/browser/progressbar.js
@@ -18,6 +18,42 @@ define((require, exports, module) => {
   const {Element, VirtualAttribute} = require('./element');
   const Component = require('omniscient');
 
+  // Animation parameters:
+  const A = 0.2;              // Zone A size (a full progress is equal to '1'
+  const B = 0.2;              // Zone B size
+  const APivot = 200;         // When to reach ~80% of zone A
+  const BPivot = 500;         // When to reach ~80% of zone B
+  const CDuration = 200;     // Time it takes to fill zone C
+  const ApproachFunc = (tMs, pivoMs) => 2 * Math.atan(tMs / pivoMs) / Math.PI;
+
+  function ComputeProgress(viewer) {
+    // Inverse tangent function: [0 - inf] -> [0 - PI/2]
+    // ApproachFunc: [time, pivot] -> [0 - 1]
+    // Pivot value is more or less when the animation seriously starts to slow down
+    let progress = 0; // value between 0 and 1
+
+    let now = performance.now();
+
+    if (viewer.get('readyState')) { // Not an empty viewer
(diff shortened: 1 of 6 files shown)


paulrouget:
Do not review. WIP.

paulrouget:
@Gozala r?

src/browser/progressbar.js
```diff
@@ -18,6 +18,42 @@ define((require, exports, module) => {
...
 
+  // Animation parameters:
+  const A = 0.2;              // Zone A size (a full progress is equal to '1'
+  const B = 0.2;              // Zone B size
+  const APivot = 200;         // When to reach ~80% of zone A
+  const BPivot = 500;         // When to reach ~80% of zone B
+  const CDuration = 200;     // Time it takes to fill zone C
+  const ApproachFunc = (tMs, pivoMs) => 2 * Math.atan(tMs / pivoMs) / Math.PI;
+
+  function ComputeProgress(viewer) {
```
Gozala: Nit: Could you please use lower leading camel case names for plain functions, and upper case leading for constructors  / factories ?

src/browser/progressbar.js
```diff
@@ -18,6 +18,42 @@ define((require, exports, module) => {
...
+    // Inverse tangent function: [0 - inf] -> [0 - PI/2]
+    // ApproachFunc: [time, pivot] -> [0 - 1]
+    // Pivot value is more or less when the animation seriously starts to slow down
+    let progress = 0; // value between 0 and 1
+
+    let now = performance.now();
+
+    if (viewer.get('readyState')) { // Not an empty viewer
```
Gozala: Suggestion: Maybe we should have a function to check if viewer isn't empty as in some places we check for `uri` field here for `readyState` I think that way we may end up causing regressions any time any part of data is changed.

src/browser/progressbar.js
```diff
@@ -26,72 +62,43 @@ define((require, exports, module) => {
...
-          progress = A_offset + B * ApproachFunc(now - connectedAt, BPivot);
-        }
-      } else {                                                                  // Zone C
-        let lastProgress = viewer.get('progress');
-        if (lastProgress < 1) {
-          progress = lastProgress + (now - before) * CSpeed;
-        } else {
-          progress = 1;
+      // Stop if loaded and had enough time to draw the final animation
+      if (viewer.get('readyState') && !viewer.get('isLoading')) {
```
Gozala: Nit: I'm not sure we may have `isLoading: true` without `readyState` that would seem odd. If we do have such case we should probably

PR #181, opened 2015-03-04
Hardcoded dashboard

gordonbrander:
Initial hardcoded dashboard. This was a WIP which has been updated. LMK if we want to rebase the commit history.

__Note__: this patch introduces a new field to the global state, so you will have to reset your session stored in localstorage.

Fixes https://github.com/mozilla/browser.html/issues/151/


src/browser/actions.js
```diff
@@ -62,6 +104,16 @@ define((require, exports, module) => {
     localStorage[`session@${version}`] = JSON.stringify(session.toJSON());
   };
 
+  const showDashboard = (dashboardCursor, tabStripCursor) => {
```
Gozala: If weather tabs trip's state is derived from dashboards, I'd rather have a something like:

```js
const isTabsTripVisible = tabStripCursor.get('isActive') || dashboardCursor.get('isActive')
```

Instead of synchronizing two fields.

gordonbrander: Makes sense

Gozala: It also I guess depends what we want behavior to be, should hiding dashboard actually trigger hiding of a tabStrip ?

src/browser/browser.js
```diff
@@ -182,8 +189,14 @@ define((require, exports, module) => {
                }
               }),
 
+      LargeTile.Deck({key: 'dashboard',
+                      className: 'dashboard',
+                      hidden: !dashboardIsActive,
+                      items: dashboardItemsCursor}),
+
```
Gozala: It could be that not even creating a `LargeTile.Deck` would be a better option, I guess depends how often we're going to be changing `dashboardItems`.

src/browser/dashboard.js
```diff
@@ -0,0 +1,30 @@
...
+    return DOM.div({key: item.get('uri'),
+             className: 'tile tile-large'}, [
+             DOM.div({key: 'tileThumbnail',
+                      className: 'tile-thumbnail',
+                      style: {backgroundImage: readBackground(item.get('image'))}}),
+             DOM.div({key: 'tileTitle',
+                      className: 'tile-title'}, null, item.get('title'))])
+  });
+
+  LargeTile.Deck = Deck(LargeTile);
```
Gozala: Suggestion: I would just call it `Dashboard` since `LargeTile.Deck` is not really descriptive IMO.

src/browser/navigation-panel.js
```diff
@@ -44,8 +44,8 @@ define((require, exports, module) => {
   const inputBindings = KeyBindings({'escape': focus});
 
 
-  const NavigationControls = Component('NavigationControls', ({inputCursor, tabStripCursor,
-                                         webViewerCursor, theme}) =>
+  const NavigationControls = Component('NavigationControls',
+    ({inputCursor, tabStripCursor, dashboardCursor, webViewerCursor, theme}) =>
```
Gozala: Suggestion: We should not pass cursors unless they affect render results, instead we should pass handler functions to trigger updating those cursors, as updating event handlers won't re-trigger components while updates to state under cursor will. At this point I suggest to either submit an issue to do it later on and put TODO: here so it's easy to find places that will require updates.

We would need to update to new version of omniscient to get swapping of event handlers without re-renders that I've contributed.

gordonbrander:
Creating a cleaned up PR

PR #182, opened 2015-03-04
Make tab switching multi-platform

Gozala:
- Remove obsolete deck actions.
- Generate diff bindings based on platform.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -94,7 +94,33 @@ define((require, exports, module) => {
   const onSelectNext = throttle(edit(selectNext), 200)
   const onSelectPrevious = throttle(edit(selectPrevious), 200);
 
-  const switchToTab = index => items => maybeActivateIndex(items, index);
+  const switchTab = (items, to) =>
+    to ? activate(select(items, to)) : items;
+
+  switchTab.toIndex = index => items => switchTab(items, items.get(index));
+  switchTab.toLast = items => switchTab(items, items.last());
+
+
+  let onTabSwitch;
+  {
+    const os = navigator.platform.startsWith('Win') ? 'windows' :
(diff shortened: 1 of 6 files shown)


src/browser/browser.js
```diff
@@ -94,7 +94,33 @@ define((require, exports, module) => {
...
+  switchTab.toIndex = index => items => switchTab(items, items.get(index));
+  switchTab.toLast = items => switchTab(items, items.last());
+
+
+  let onTabSwitch;
+  {
+    const os = navigator.platform.startsWith('Win') ? 'windows' :
+               navigator.platform.startsWith('Mac') ? 'osx' :
+               navigator.platform.startsWith('Linux') ? 'linux' :
+               '';
```
fabricedesre: we already do that in src/browser/keyboard.js and src/browser/actions.js. Looks like we should move it to a common place.

Gozala: Yeah I was thinking about that too, I'll send a new patch to do so.

src/os.js
```diff
@@ -0,0 +1,12 @@
+// Subset of `os` module from node.js and io.js:
+// https://iojs.org/api/os.html
+define((require, exports, module) => {
+  const platform = navigator.platform.startsWith('Win') ? 'win32' :
```
fabricedesre: nit: I would not say win32, as it's also valid for 64 bits windows.

src/os.js
```diff
@@ -0,0 +1,12 @@
...
+                   navigator.platform.startsWith('Mac') ? 'darwin' :
+                   navigator.platform.startsWith('Linux') ? 'linux' :
+                   navigator.platform.startsWith('FreeBSD') ? 'freebsd' :
```
fabricedesre: ;) did you test that? What about BeOS!?

Issue #123, opened 2015-02-19
ctl+tab: tab shade is shown but location bar is not


gordonbrander:
`ctl+tab` when location bar is scrolled off screen results in tab shade without location bar. Before we can fix this issue, we will have to solve the problem of showing the location bar without moving the content position (same issue as https://github.com/mozilla/browser.html/issues/96). cc/ @paulrouget

gordonbrander:
Out of date.

Issue #96, opened 2015-02-18
Show title bar when mousing over hotzone


gordonbrander:
When you mouse toward the title bar and pass over the hot zone, the title bar should slide down. Content visual position should not change.

The hotzone should be 100% the width of the window. Let's start with a hotzone that's 200px high and keep it easy to tweak so we can tune for the right size.

paulrouget:
I will take care of that. I'll try to hack something to synchronize chrome and content scrolling to translate the titlebar back in the viewport.

paulrouget:
This is what I tried:
- expose scrolling to the browser api (patch for the record: https://gist.github.com/paulrouget/e5acbe78d0e4cde5e992 I didn't manage to make it synchronous yet)
- scroll `<main>` by `+x` and content by `-x`

It flickers pretty badly, and of course, animation is super bad.

No surprise, the call is async and go through IPC.

I could make it sync, but that would change that we have to go through `main process -> (DOM call) -> child process -> scroll -> compositor (main process)`.

I see 2 options:
- make it possible at the compositor level to scroll `<main>` and keep the position of the content
- find a different mechanism to hide/show the navbar, something not based on scrollgrab + snap points

paulrouget:
So I've been experimenting with different approaches. Sadly, to get anything to work properly, we need to track the scroll position of the content, which only works if APZC is enabled. But APZC is really broken on Mac.

No matter what we do, it won't be an optimal solution.

So we can, either, give up on this feature for now until the platform allows us to do that (see http://bugzil.la/1138896, http://bugzil.la/1137872 and https://groups.google.com/forum/?fromgroups#!topic/mozilla.dev.platform/K_NhNWVKOPI). Also, waiting for APZC will make our life easier.

Or we can build a temporary patch.

Here, 2 approaches:

1. do not base this mechanism on scroll position. Show/hide the navbar if we're in the hotzone or not. Scrolling won't hide the navbar.
2. we do implement the double-scroll thing. Every time the navbar shows up, the animation will be choppy. This will also require a platform patch (we have it, but it will need reviews, not sure it will ever land).

gordonbrander:
3. A temporary fix where we render 2 title bars: one at the top, one `position: fixed` above the screen. We show the `position: fixed` title bar when mousing into the hotzone. This obviously will cover `position: fixed` elements on the page, but could be a good temporary approach while platform gets fixed.

gordonbrander:
[...]

PR #188, opened 2015-03-05
Bring back navbar if mouse reaches hotzone

paulrouget:
See issue #96.

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -115,6 +115,26 @@ define((require, exports, module) => {
     return session;
   };
 
+  const showNavbar = () => {
+    // smoothScrollBy requires a platform patch to work.
+    // We don't want to scroll the navbar if we can't
+    // scroll the iframe later.
+    if (HTMLIFrameElement.prototype.smoothScrollBy) {
+      const main = document.querySelector('main');
+      const scrollTop = main.scrollTop;
+      if (scrollTop != 0) {
+        const iframe = document.querySelector('iframe:not([hidden=true])')
+        if (iframe) {
+          iframe.addEventListener('mozbrowseron-scroll-in-rfa', function scroll() {
+            iframe.removeEventListener('mozbrowseron-scroll-in-rfa', scroll);
+            main.scrollBy({top: -scrollTop, behavior: 'smooth'});
+          });
+          iframe.smoothScrollBy(0, scrollTop); // smooth scroll over IPC
+        }
+      }
+    }
+  }
+
   // Exports:
 
   exports.makeSearchURL = makeSearchURL;
@@ -125,5 +145,6 @@ define((require, exports, module) => {
   exports.readSession = readSession;
   exports.writeSession = writeSession;
   exports.sendEventToChrome = sendEventToChrome;
+  exports.showNavbar = showNavbar;
 
 });


paulrouget:
@Gozala Can you review that? You need a special to make that work. Get the "syncScroll" build to try it out. As discussed in #96, we know it's a not smooth, but it requires some platform work to make it better.

As you can see, the scroll operations are called in an isolated function. I could have done it in the render function. I tried that at first, but it requires to reach the main DOM node from the iframe to check the initial scroll position, or to store the scroll position in the state, but that would require a scroll event listener, and I'd like to avoid that. Let me know if you think of a better way to do that.

Fixes #123 and #96

Gozala:
> As you can see, the scroll operations are called in an isolated function. I could have done it in the render function. I tried that at first, but it requires to reach the main DOM node from the iframe to check the initial scroll position, or to store the scroll position in the state, but that would require a scroll event listener, and I'd like to avoid that. Let me know if you think of a better way to do that.

It's a second occasion we need to stimulate updates using request animation frame, which is fine but I'd like to find a better / less hacky way to deal with that. I need to think about it though as I don't have anything specific in mind yet.

## Just thinking out loud

One pattern I do notice though is that we do emulate events that are either too slow to listen or don't note exist at all. Maybe we could use that `VirtualEvent` API that I have added recently to deal with that:

https://github.com/mozilla/browser.html/blob/master/src/browser/element.js#L219-L244

And here is how it's used:
https://github.com/mozilla/browser.html/blob/b20384e27b67f93ba9c14f8e945bb23a63f1500f/src/browser/iframe.js#L84-L100

Although I do see that in both cases we need access to a state to decide if we need to register another RFA.

Maybe we could have something like `onAnimationFrame` `VirtualEvent` that will keep firing on animation frames unless handler is unassigned, that way decisions about the state could be made in `render` where we do have all the state and based of it we could set `onAnimationFrame` event handler or not.

Although we probably need to modify `VirtualEvent` implementation like this line:
https://github.com/mozilla/browser.html/blob/master/src/browser/element.js#L233 to pass `this` instead of `this.handleEvent.bind(this)` so that on animation frame we could check if `this.handler` is set in order to decide if next inaction frame should be scheduled, or maybe we could expose another hook to achieve same.

I think this would allow you to update scroll position in the state only when it matters & at the animation frame, that than you'd be able to use in a `render` of the iframe to trigger a scroll.

Gozala:
P.S.: Those were just some thoughts, I'm not sure if it's worth spending time cleaning all this up right now, it could be that'll be better to do such cleanups afterwards.

paulrouget:
I'm more concerned about the fact I have to use `querySelector`, making this code dependent on components internal implementation.

Gozala:
> I'm more concerned about the fact I have to use querySelector, making this code dependent on components internal implementation.

I think suggestion I made would address that as well.

paulrouget:
Sorry, I don't understand. Forget about the RFA. This has nothing to do with the issue I'm facing here.

Let me explain better: when we enter the hotzone, the `iframe` needs to scroll only if `main.scrollTop != 0`. So when the mouse enters the hotzone, I set `mouseInHotzone` to `true`. Then when `main` is rendered, if `main.scrollTop != 0`, I scroll it back to the top. To compensate that, I need to scroll the content a tiny bit. So when it's the turn of the `iframe` to be rendered, I need to know if `main` required a scroll or not.

To do so, I have 3 options:
1. go check main's scroll position. If it's not `0`, then I know I need to scroll the content as well.
2. read a state that would have been updated by `main.onScroll()`
3. when rendering `main`, if a scroll was required, set a property `scrollContentRequired` in the state

1. is simple a very straight forward. But requires knowing the implementation of another component.
2. is over-complicated. Add a scroll listener sounds too much. Especially when 1. is possible.
3. … I don't know where to do that. In `componentDidUpdate`? Is that synchronous with the `render` function? Also, I would need to reset the `scrollContentRequired`, where would I do that? And wouldn't that require 3 rendering?

I really want to go for 1 (which is what happens in the PR), because it's really simple and straight forward.

paulrouget:
@Gozala ping?

src/browser/browser.js
```diff
@@ -160,9 +165,16 @@ define((require, exports, module) => {
 
     const theme = Browser.readTheme(activeWebViewerCursor);
 
+    const hotZone = immutableState.get('hotZone');
+
+    if (isTabStripVisible) {
+      showNavbar();
+    }
```
Gozala: React really advices against touching DOM during render phase. I think it would be much better if you could handle DOM updates in one of the virtual attributes like you already do with `mouseInHotZone`. In fact I don't really understand why it can't be something like:

```js
Main({
  navbarVisible: isTabStripVisible || hotZone.get('mouseIn'),
  //...
})
```

That would make more cleaner and also would not go against React's don't touch DOM during render phase.

src/browser/browser.js
```diff
@@ -36,6 +36,11 @@ define((require, exports, module) => {
     scrollGrab: VirtualAttribute((node, current, past) => {
       node.scrollgrab = current;
     }),
+    mouseInHotzone: VirtualAttribute((node, current, past) => {
+      if (current && !past) {
+        showNavbar();
```
Gozala: BTW `node` here is the `main` so you could pass it into `showNavBar` function to avoid using `querySelector`.

src/browser/actions.js
```diff
@@ -59,6 +59,25 @@ define((require, exports, module) => {
...
+    // smoothScrollBy requires a platform patch to work.
+    // We don't want to scroll the navbar if we can't
+    // scroll the iframe later.
+    if (HTMLIFrameElement.prototype.smoothScrollBy) {
+      const main = document.querySelector('main');
+      const scrollTop = main.scrollTop;
+      if (scrollTop != 0) {
+        requestAnimationFrame(() => {
+          main.scrollBy({top: -scrollTop, behavior: 'smooth'});
+        });
```
Gozala: Why do we scroll main on the next frame and iframe within this frame ?

Gozala:
Ok here is my view on this matter. While I understand that current solution is simple and works I can't help not feeling bad about it. Since it couples two unrelated components and encodes knowledge of which iframe is active into the code that should not care about it. 

The second option on the other hand seems clear as state will have `scrollOffset` and whenever `isTabStripVisible || hotZone.get('mouseIn')` is true active `webViewer` and `main` would just get an attribute that will be reflected then on dom in form of virtual attributes.

As far as I get the main reason you don't like this option is the fact that it requires to listen to scroll events, if that is so, I think the best approach would be to get around this so we don't have to listen to scroll events. I think that should be very much doable, we can define `VirtualEvent` say `onScrollOffset` (that does not need to go through DOM events) which we can invoke only if virtual `isNavbarVisible` attribute on `main` is set to `true` and `node.scrollTop != 0` with details like `node.scrollTop` as a side effect state will be updated and relevant components will reflect that by smooth scrolling.

paulrouget:
My comment from issue #331:

> If we want to go for a not-perfect-but-good-enough solution, I would recommend a different approach: we could send an async message to the child process to request a sync message from the child process during a RFA, we smooth scroll the child process after the sync message, and smooth scroll the main process when we get the sync message.
> I think this will ensure that both smooth scrolls are part of the same transactions.
> There's still the IPC call from the child process that request the smoothScroll that might get some lag.
> I don't know enough about how scroll operations are triggered from the child process (with or without async scroll), especially with smooth scroll operations.
> We should try at least.

I tried and it appears to work. It requires APZC though.

Gozala:
This pull request is obsolete at this point.

PR #192, opened 2015-03-05
Update to omniscient 3.1.0.

Gozala:
With latest omniscient we can avoid re-triggering component by passing event handlers as a second argument:

https://github.com/mozilla/browser.html/wiki/Design-F.A.Q.#q-how-do-update-state-that-i-dont-need-to-render-


src/browser/browser.js
```diff
@@ -210,12 +211,14 @@ define((require, exports, module) => {
...
               }),
 
-      WebViewer.Deck({key: 'web-viewers',
-                      className: 'iframes',
-                      items: webViewersCursor,
-                      onClose: item => webViewersCursor.update(closeTab(item)),
-                      onOpen: item => webViewersCursor.update(addTab(item))
-                     })
+      WebViewer.Deck({
+        key: 'web-viewers',
```
gordonbrander: Comment: FWIW, I prefer this style of indentation in C-style languages because it less frequently runs up against the 80 char limit.

gordonbrander:
r+

PR #194, opened 2015-03-06
Take advantage of caching API to avoid extra computations.

Gozala:
https://github.com/mozilla/browser.html/wiki/Design-F.A.Q.#q-how-to-avoid-recomputing-derived-state-on-each-update-


gordonbrander:
R+

PR #195, opened 2015-03-06
Implement tab ordering by switch time.

Gozala:
# Work in progress

- [x] Order tabs by recent switch
- [x] navigation between tabs by switch recency.

**P.S.:** This change depends on new fields in data model so in order for you tabs to start render in right order you'd need to activate each one or reset session with `cmd shift backspace`

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -70,8 +70,6 @@ define((require, exports, module) => {
     tabStripCursor.set('isActive', true);
   exports.hideTabStrip = tabStripCursor =>
     tabStripCursor.set('isActive', false);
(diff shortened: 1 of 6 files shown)


gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/196

gordonbrander:
@Gozala  one small jank issue:

1. Open tab shade, select tab with mouse
2. Click
3. Mouse out

As the mouse moves out of the tab shade before it disappears you can see a flash when the tabs get re-rendered (the tab you selected moving to the front). Then it disappears. It's fast enough that you can't see what's happening and feels like a lot of visual noise.

We should hide then re-render.

PR #208, opened 2015-03-10
Add hardcoded dashboard

gordonbrander:
Squashed and cleaned up dashboard commits. Fixes https://github.com/mozilla/browser.html/issues/151.

### DIFF
--- a/css/dashboard.css
+++ b/css/dashboard.css
@@ -1,24 +1,32 @@
 .dashboard {
-  flex: 0 0 800px;
   align-content: center;
+  flex: 0 0 100%;
   flex-flow: row wrap;
   /* The top margin places the dashboard below the tabstrip, the bottom margin
   visually centers the tiles vertically. We need the bottom margin because we're
   using flexbox to align the tiles, but the tiles themselves are not visually
   even, since they have extra space below for the title. */
-  margin: 100px auto 30px;
+  margin: 0 auto;
   width: 800px;
+  padding: 100px 0 30px;
 }
 
 .tile-large {
   align-items: center;
+  /* Setting the flex property on tiles is important to ensure that large titles
+  don't blow up the grid. */
+  flex: 0 0 200px;
   flex-direction: column;
+  /* Setting the height allows us to maintain extra space beneath title */
+  height: 150px;
+  padding-top: 20px;
 }
 
 .tile-large .tile-thumbnail {
   background: #fff;
   background-size: cover;
   border-radius: 12px;
+  border: 1px solid rgba(0,0,0,0.1);
   height: 100px;
   transition: transform 80ms ease-out;
   will-change: transform;
@@ -34,7 +42,6 @@
   font-weight: bold;
   font-size: 12px;
   opacity: 0;
-  height: 30px;
   padding: 15px 0 0;
   transition: opacity 160ms ease-out;
   white-space: nowrap;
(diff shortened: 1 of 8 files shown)


paulrouget:
I'll review that.

css/dashboard.css
```diff
@@ -0,0 +1,51 @@
...
+  flex: 0 0 800px;
+  align-content: center;
+  flex-flow: row wrap;
+  /* The top margin places the dashboard below the tabstrip, the bottom margin
+  visually centers the tiles vertically. We need the bottom margin because we're
+  using flexbox to align the tiles, but the tiles themselves are not visually
+  even, since they have extra space below for the title. */
+  margin: 100px auto 30px;
+  width: 800px;
+}
```
paulrouget: Make it responsive:
```
  flex-wrap: wrap;
  margin: 100px auto 30px;
  max-width: 800px;
  align-content: flex-start;
  justify-content: center;
```

paulrouget:
Let's render the dashboard at the browser level, no in the webviewer. The dashboard would be re-rendered everytime. If we render it once, and hide it behind the webviewer deck, it should work.

gordonbrander:
@paulrouget I tried that first, but it requires a lot of state rewiring. This was the least invasive. If we want to do it that way, I would like to get @Gozala's help.

paulrouget:
>  but it requires a lot of state rewiring

How so? Can I help?

gordonbrander:
Wait a minute... I'm a dope. I can just add `hidden` to the rendered dashboard by checking the active webviewer and seeing if it has a URL. Ok. Will update commit tomorrow AM.

css/dashboard.css
gordonbrander: These styles don't do what is intended... they center an uneven number of tiles (should be left-justified) and also align tiles to the top of the screen space, rather than centering them in space below tab strip.

Will update to be responsive.

gordonbrander: @paulrouget Let's do the responsiveness in a separate update. Rather than reflowing tiles, I want to physically shrink the tiles using a media query so the grid is always 4x3.

It is important that the grid always stay the same if we plan to use it for bookmarks.

gordonbrander:
Ok, I've added a responsive breakpoint for `< 800px`. @paulrouget please review.

src/browser/browser.js
```diff
@@ -199,24 +205,32 @@ define((require, exports, module) => {
...
+        key: 'tabstripkillzone',
+        className: ClassSet({
+          tabstripkillzone: true,
+          'tabstripkillzone-hidden': !isTabStripVisible || isDashboardActive
+        }),
+        onMouseEnter: event => hideTabStrip(tabStripCursor)
+      }),
+      Dashboard({
+        items: dashboardItems,
+        hidden: activeWebViewerCursor.get('uri') !== null
```
paulrouget: Why don't you use `isDashboardActive`?

src/browser/browser.js
```diff
@@ -199,24 +205,32 @@ define((require, exports, module) => {
...
+      }, {
+        onOpen: uri => activeWebViewerCursor.set('uri', uri)
+      }),
       WebViewer.Deck({
         key: 'web-viewers',
         className: 'iframes',
+        hidden: activeWebViewerCursor.get('uri') === null,
```
paulrouget: same.

src/browser/browser.js
gordonbrander: Yup, you're right

css/theme.css
```diff
@@ -75,10 +75,14 @@ iframe {
 }
 
 .iframes {
-  flex-grow: 1;
+  flex: 0 0 100%;
   scroll-snap-coordinate: 0 0;
 }
```
paulrouget: Why do you change that?

css/theme.css
```diff
@@ -75,10 +75,14 @@ iframe {
...
 
+.iframes-frame {
+  flex: 0 0 100%;
+}
+
```
paulrouget: Why? I'm surprised you don't want a `flex-grow:1` here.

PR #217, opened 2015-03-11
Store user input at the webviewer level

paulrouget:
It makes the awesomebar behave better, and also how Firefox Desktop works.


paulrouget:
@Gozala Please review this.

paulrouget:
`gozala 08:20: seems good `

PR #88, opened 2015-02-18
Fallback to thumbnail #83

Gozala:
# Work in progress

PR #220, opened 2015-03-11
Add thumbnails for tabs without tiles.

Gozala:
#88


src/browser/web-viewer.js
```diff
@@ -101,9 +118,31 @@ define((require, exports, module) => {
...
+    request.onload = event => resolve(request.response);
+    request.onerror = event => reject()
+  });
+
+  WebViewer.onThumbnailChanged = webViewerCursor => blob =>
+    webViewerCursor.set('thumbnail', URL.createObjectURL(blob));
+
+  const fetchScreenshot = iframe => new Promise((resolve, reject) => {
+    const request = iframe.getScreenshot(100, 62.5, 'image/png');
+    request.onsuccess = event => resolve(request.result);
```
paulrouget: Isn't it a promise? (`.then`)

paulrouget:
Not sure why your code is not working, but I've tried that locally:

`iframe.getScreenshot(100,100).then(s => console.log(URL.createObjectURL(s)))`

…and it works even on hidden iframes (didn't even `setVisibility(true)`).

src/browser/page-switch.js
```diff
@@ -31,17 +29,26 @@ define((require, exports, moudle) => {
...
         className: 'tab-thumbnail',
-        style: {backgroundImage: readThumbnailURI(webViewerCursor.get('location'))},
-      }),
+      }, [
+        DOM.img({
+          key: 'image',
+          className: 'image',
+          src: thumbnail,
+          alt: '',
+          onLoad: event => URL.revokeObjectURL(event.target.src),
```
paulrouget: That's suspicious. This will revoke the URL you just loaded and next gc will discard your image.

src/browser/web-viewer.js
```diff
@@ -101,9 +118,31 @@ define((require, exports, module) => {
...
+    request.send();
+    request.onload = event => resolve(request.response);
+    request.onerror = event => reject()
+  });
+
+  WebViewer.onThumbnailChanged = webViewerCursor => blob =>
+    webViewerCursor.set('thumbnail', URL.createObjectURL(blob));
+
+  const fetchScreenshot = iframe => new Promise((resolve, reject) => {
+    const request = iframe.getScreenshot(100, 62.5, 'image/png');
```
paulrouget: Isn't `request` a promise already? Test `request.then`.

Gozala: > Isn't request a promise already? Test request.then.

I wish, it's [DOMRequest](https://developer.mozilla.org/en-US/docs/Web/API/DOMRequest)

paulrouget: It's a promise here. WTF.

src/browser/page-switch.js
Gozala: As far as I understand that's the way blob urls are supposed to be used. Sure blob will be gc-ed but we don't care image will be rendered and isn't going to go away unless it's remove from DOM at which point we don't care.

Gozala: In fact if you google about blob urls you'll find stuff like this
http://blogs.adobe.com/webplatform/2012/01/17/displaying-xhr-downloaded-images-using-the-file-api/

Gozala:
So I had to rewrite the way tab ordering works as otherwise I was hitting these bugs in cursor implementation:

https://github.com/facebook/immutable-js/issues/370
https://github.com/omniscientjs/omniscient/issues/89
https://github.com/omniscientjs/omniscient/issues/91

Now `order` is encoded into a data structure, and css `order` is used to render tabs in the desired order instead of re-arranging items in the data structure itself. This actually has another benefit of not having to re-render or do that much dom updates. Although primary goal was to fix issue when you switch tabs and thumbnail for a wrong tab would update.

css/tabstrip.css
```diff
@@ -47,6 +47,7 @@
 }
 
 .tab-thumbnail {
+  position: relative;
   background: white;
   background-size: cover;
```
paulrouget: If we use `` now, you can remove the background related CSS code.

css/tabstrip.css
```diff
@@ -56,8 +57,19 @@
   transform: scale(0.8);
   transition: transform 150ms ease;
   width: 100px;
+  overflow: hidden;
+
 }
 
+.tab-thumbnail .image {
+  position: absolute;
```
paulrouget: Why?

css/tabstrip.css
```diff
@@ -56,8 +57,19 @@
...
+  width: inherit;
+  height: inherit;
+  -moz-force-broken-image-icon: 0;
```
paulrouget: It is 0 by default. If you see the broken image icon, we need to set `browser.display.show_image_placeholders` to false in graphene.

src/browser/web-viewer.js
```diff
@@ -30,7 +37,10 @@ define((require, exports, module) => {
       isRemote: true,
       allowFullScreen: true,
 
-      isVisible: isActive(webViewerCursor) || isSelected(webViewerCursor),
+      isVisible: isActive(webViewerCursor) ||
+                 isSelected(webViewerCursor) ||
+                 !webViewerCursor.get('thumbnail'),
```
paulrouget: Do you really have to do that? It doesn't work if you don't set the iframe visible?

src/browser/web-viewer.js
```diff
@@ -60,6 +70,48 @@ define((require, exports, module) => {
     })
   });
 
+
+  const fetchScreenshot = iframe =>
+    fromDOMRequest(iframe.getScreenshot(100, 62.5, 'image/png'));
```
paulrouget: Multiply that by `devicePixelRatio`.

Gozala: > Multiply that by devicePixelRatio.

How do I find it out ?

paulrouget: `window.devicePixelRatio`

src/browser/web-viewer.js
Gozala: > Do you really have to do that? It doesn't work if you don't set the iframe visible?

Looks like it works without making iframe visible as well.

paulrouget:
This totally breaks tab navigation, I can't close tabs anymore, and ctrl-shit-tab doesn't appear to work correctly.

src/browser/page-switch.js
```diff
@@ -31,21 +30,35 @@ define((require, exports, moudle) => {
...
         key: 'thumbnail',
         className: 'tab-thumbnail',
-        style: {backgroundImage: readThumbnailURI(webViewerCursor.get('location'))},
-      }),
+      }, [
+        DOM.img({
+          key: 'image',
+          src: thumbnail,
+          alt: '',
+          onLoad: event => URL.revokeObjectURL(event.target.src)
```
paulrouget: 1. you store thumbnail in the state. That will break on restart, as the blobs are not stored across sessions
2. you revoke the URL after load, making re-using the blob impossible even in the same session

Gozala: > you store thumbnail in the state. That will break on restart, as the blobs are not stored across sessions

I don' really store blobs I store blob uris (see https://github.com/mozilla/browser.html/pull/220/files#diff-ac64a61ab532164b9e3f853bfb7ee2d7R201), but yeah they won't work across the sessions. I think for that we need to implement / revive proper database like pouchdb. But I think we were going to defer that for now.

> you revoke the URL after load, making re-using the blob impossible even in the same session

Sure but we do not re-render those images ever with in the same session, if tab is gone then all associated data is gone anyway. I could store actual blob in the state and create url during render and revoke after it loads but I don't find it to be any better.

src/browser/web-viewer.js
```diff
@@ -100,10 +155,15 @@ define((require, exports, module) => {
...
-    webViewerCursor.merge(Object.assign({location: event.detail,
-                                        userInput: event.detail},
-                                        getHardcodedColors(event.detail)));
+    webViewerCursor.merge({location: event.detail,
+                           userInput: event.detail}).
+                    merge(getHardcodedColors(event.detail));
+
+    requestThumbnail(event.target).
+      then(WebViewer.onThumbnailChanged(webViewerCursor),
+           e => console.error(e));
```
paulrouget: I see this error in the console a lot.

> mozbrowserlocationchange { target: <iframe.iframes-frame.webviewer.contentoverflows>, isTrusted: true, detail: "https://duckduckgo.com/?q=hackernews", eventPhase: 0, bubbles: true, cancelable: false, defaultPrevented: false, timeStamp: 1426128248631231, originalTarget: <iframe.iframes-frame.webviewer.contentoverflows>, explicitOriginalTarget: <iframe.iframes-frame.webviewer.contentoverflows>, NONE: 0 }

css/tabstrip.css
```diff
@@ -56,8 +55,10 @@
   transform: scale(0.8);
   transition: transform 150ms ease;
   width: 100px;
+  overflow: hidden;
```
paulrouget: This is not supposed to be required. The image is supposed to be the exact size of its container.

src/browser/page-switch.js
paulrouget: You're right.

src/browser/web-viewer.js
```diff
@@ -60,6 +69,50 @@ define((require, exports, module) => {
...
   });
 
+
+  const fetchScreenshot = iframe =>
+    fromDOMRequest(iframe.getScreenshot(100 * devicePixelRatio,
+                                        62.5 * devicePixelRatio,
+                                        'image/png'));
+
+  const fetchThumbnail = uri => new Promise((resolve, reject) => {
+    const request = new XMLHttpRequest();
```
paulrouget: Just add a comment here to say that this bad, we know it's bad, and we'll fix that later.

src/browser/web-viewer.js
paulrouget: I think this is the abort promise that fails (as expected).

paulrouget:
r+ with the 3 last comments addressed.

Gozala:
> This totally breaks tab navigation, I can't close tabs anymore, and ctrl-shit-tab doesn't appear to work correctly.

Do you still have this ?

src/browser/web-viewer.js
Gozala: > I think this is the abort promise that fails (as expected).

Yes while debugging I was looking at when promise would fail and why, I'll remove that.

PR #230, opened 2015-03-12
navigate through the awesomebar with keys

### LINKED ISSUE #225: Add support for "arrow down" to awesomebar results
Currently you have to click.

cc @paulrouget

paulrouget:
Fix #225

### DIFF
--- a/src/browser/awesomebar.js
+++ b/src/browser/awesomebar.js
@@ -11,42 +11,15 @@ define((require, exports, module) => {
   const {DOM} = require('react')
   const {fromJS} = require('immutable');
   const Component = require('omniscient');
-  const {Element, Event} = require('./element');
+  const {Element} = require('./element');
+  const ClassSet = require('./util/class-set');
 
   const Awesomebar = Component('Awesomebar', (
     {suggestionsCursor, theme},
     {onOpen}) => {
 
-    let suggestions = [];
-
-    let search = suggestionsCursor.get('search').toArray();
-    let history = suggestionsCursor.get('history').toArray();
-
-    // searchCount + historyCount == 6
-    let searchCount = Math.min(search.length,
-                               MAX_RESULTS - Math.min(MAX_RESULTS / 2, history.length));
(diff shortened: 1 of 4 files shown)


paulrouget:
@Gozala r?

src/browser/awesomebar.js
```diff
@@ -11,42 +11,15 @@ define((require, exports, module) => {
...
-      let href = history[i].get('href');
-      let text = history[i].get('text');
-      suggestions.push(DOM.p({
-        className: 'suggestion history',
-        key: 'suggestionHistory' + i,
-        href,
-      }, text));
-    }
+    const selectedIndex = suggestionsCursor.get('selectedIndex');
+    const list = suggestionsCursor.get('list').toJSON();
```
Gozala: Nit: You can call `.map()` on Immutable list, no need to transform to JSON.

src/browser/navigation-panel.js
```diff
@@ -45,17 +46,76 @@ define((require, exports, module) => {
...
+    const size = suggestionsCursor.get('list').size;
+    let value = suggestionsCursor.get('selectedIndex') + delta;
+    if (value >= size) {
+      value = -1;
+    }
+    value = Math.max(-1, value);
+    suggestionsCursor.set('selectedIndex', value);
+  }
+
+  const onInputKeyDown = ({event, inputCursor, webViewerCursor, suggestionsCursor}) => {
```
Gozala: Can you please use `KeyBindings` instead as it does basically what you do here.

Gozala:
As a general note, you are abusing `fromJS` in most cases you don't actually need it. Also use of Immutable (Map, List, Set) constructors tends to be faster.

Gozala:
Overall this looks alright, but I would very much like to leverage existing `KeyBindings` for defining keybindings as it has more declarative interface and + all issues will be scoped in single location.

That being said I don't mind doing it as a followup pull.

Issue #234, opened 2015-03-12
Awesomebar keyboard stuff should use `Keybindings`


Gozala:
This is followup for #230

Gozala:
Fixed by #236

PR #236, opened 2015-03-12
Use keybindings in awesomebar.

Gozala:
Fix for #234

### DIFF
--- a/src/browser/keyboard.js
+++ b/src/browser/keyboard.js
@@ -75,6 +75,8 @@ define((require, exports, module) => {
 
         if (binding) {
           binding(...args);
+          event.preventDefault();
+          event.stopPropagation();
         }
       }
       return event;
--- a/src/browser/navigation-panel.js
+++ b/src/browser/navigation-panel.js
@@ -11,12 +11,12 @@ define((require, exports, module) => {
   const Component = require('omniscient');
   const {InputField} = require('./editable');
   const {Element} = require('./element');
-  const {navigateTo, showTabStrip, blur, focus} = require('./actions');
+  const {navigateTo, showTabStrip, blur, focus, select} = require('./actions');
   const {KeyBindings} = require('./keyboard');
   const url = require('./util/url');
   const {ProgressBar} = require('./progressbar');
   const ClassSet = require('./util/class-set');
-  const {throttle} = require('lang/functional');
+  const {throttle, compose, arity} = require('lang/functional');
   let {computeSuggestions, resetSuggestions} = require('./awesomebar');
 
   computeSuggestions = throttle(computeSuggestions, 200);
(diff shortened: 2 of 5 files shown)


src/browser/keyboard.js
```diff
@@ -75,6 +75,8 @@ define((require, exports, module) => {
 
         if (binding) {
           binding(...args);
+          event.preventDefault();
+          event.stopPropagation();
```
paulrouget: That's a bit violent. Can this be handled in the callback?

src/browser/navigation-panel.js
```diff
@@ -46,78 +46,80 @@ define((require, exports, module) => {
...
-  const updateSuggestionsSelection = (delta, suggestionsCursor, webViewerCursor) => {
-    const size = suggestionsCursor.get('list').size;
-    let value = suggestionsCursor.get('selectedIndex') + delta;
-    if (value >= size) {
-      value = -1;
+  // Clears selection in the suggestions
+  const unselect = suggestions =>
+    suggestions.set('selectedIndex', -1);
+
+  // Moves selecetion from current index with give `delta`.
```
paulrouget: typo.

src/browser/navigation-panel.js
```diff
@@ -46,78 +46,80 @@ define((require, exports, module) => {
...
+
+  // Selects next / previous suggestion.
+  const previous = move(-1);
+  const next = move(1);
+
+  // Returns currently selected suggestion.
+  const selected = suggestions => {
+    // When the user starts doing something that is not navigating
+    // through the suggestions, if a suggestion was selected, we
+    // commit it as a userInput, then the suer can start editing it
```
paulrouget: type (my fault)

src/browser/navigation-panel.js
```diff
@@ -128,10 +130,11 @@ define((require, exports, module) => {
         className: 'urlinput',
         style: theme.urlInput,
         placeholder: 'Search or enter address',
-        value: inputValue,
+        value: webViewerCursor.get('userInput'),
```
paulrouget: No. You lose the ability to come back to no selection at all. We need to be able to do that:
```
type "a"
See "amazon.com, apple.com"
key down
See "amazon.com" in the input
key up
Back to "a" in the urlbar
```

Gozala: Oh I see, makes sense I'll fix that then.

src/browser/keyboard.js
Gozala: > That's a bit violent. Can this be handled in the callback?

At the moment non of our keybindings code needs events to be propagated, how about we keep it and change when / if we need it ?

src/browser/navigation-panel.js
```diff
@@ -46,78 +46,61 @@ define((require, exports, module) => {
...
+    }));
+
+  // Selects next / previous suggestion.
+  const selectPrevious = selectRelative(-1);
+  const selectNext = selectRelative(1);
+
+  // Returns currently selected suggestion or void if there's none.
+  const selected = suggestions => {
+    const index = suggestions.get('selectedIndex');
+    return index >= 0 ? suggestions.getIn(['list', index, 'text']) : void(0);
```
Gozala: Please not that while `x.get(a).get(b)` may throw `x.getIn([a, b])` won't it will just short circuit on with `undefined` if somewhere along the path is, there for I removed try catch.

paulrouget:
r+

PR #242, opened 2015-03-13
Wallpapers

gordonbrander:
- [x] Theme dashboard
- [x] Theme navigation
- [x] Theme awesomebar
- [x] Update dashboard wallpaper and navigation on swatch click
- [x] Configure real themes
- [x] Retain chosen theme after refresh


gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/214

gordonbrander:
@Gozala r?

css/dashboard.css
```diff
@@ -1,14 +1,24 @@
 .dashboard {
+  background-position: center center;
+  background-size: cover;
+  align-items: center;
+  flex: 1 0 auto;
+  flex-flow: column nowrap;
+  color: #555;
+  justify-items: center;
```
paulrouget: This property doesn't exist.

css/awesomebar.css
```diff
@@ -34,29 +34,29 @@
   line-height: 38px;
   vertical-align: center;
```
paulrouget: s/center/middle (… I wrote that :/ )

css/dashboard.css
```diff
@@ -50,8 +59,26 @@
   opacity: 1;
 }
 
+.wallpaper-swatches {
+  flex-flow: row nowrap;
```
paulrouget: I would prefer if we could avoid shorthands for flex properties. I'm not familiar enough with them yet.

src/browser/dashboard/actions.js
```diff
@@ -0,0 +1,137 @@
...
+    );
+  }
+
+  // Exports:
+
+  exports.getDashboardThemePatch = getDashboardThemePatch;
+  exports.getWallpaperSwatches = getWallpaperSwatches;
+  exports.readDashboardTheme = readDashboardTheme;
+  exports.readDashboardNavigationTheme = readDashboardNavigationTheme;
+  exports.init = init;
```
paulrouget: All the methods use "dashboard" in their names. Do the same thing with `init`.

src/browser/dashboard/actions.js
```diff
@@ -0,0 +1,137 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define((require, exports, module) => {
+  'use strict';
+
+  const {expandCustomTheme} = require('../theme.js');
+
+  // Like `array.reduce` but for objects.
```
paulrouget: `Object.keys(object).reduce()`

src/browser/dashboard/actions.js
```diff
@@ -0,0 +1,137 @@
...
+  const reducekv = (object, step, value) => {
+    for (let key in object) {
+      value = step(value, key, object[key]);
```
paulrouget: What about using `Object.keys(object).reduce()`?

src/browser/actions.js
```diff
@@ -81,7 +82,7 @@ define((require, exports, module) => {
     // once #239 is resolved.
     input: {value: '', isFocused: false},
     tabStrip: {isActive: false},
-    dashboard: {items: dashboardItems},
+    dashboard: initDashboard({items: dashboardItems}),
```
paulrouget: What about:
```
dashboard: {
  items: dashboardItems,
  theme: getDashboardThemePatch('default')
}
```
and forget about the `init` function?

src/browser/browser.js
```diff
@@ -165,6 +167,7 @@ define((require, exports, module) => {
     const rfaCursor = immutableState.cursor('rfa');
 
     const dashboard = immutableState.get('dashboard');
+    const dashboardCursor = immutableState.cursor('dashboard');
```
paulrouget: Maybe you want to use the cursor everywhere, and kill `dashboard`.

src/browser/dashboard.js
```diff
@@ -7,12 +7,27 @@ define((require, exports, module) => {
...
-    ({key, uri, image, title}, {onOpen}) =>
+  const List = (Item, a2b) => Component('List', (options, handlers) =>
+    DOM.div(options, options.items.map(options =>
+      Item(a2b(options), handlers))));
+
+  const WallpaperSwatch = ({key, backgroundColor}, {onWallpaperChange}) =>
+    DOM.div({
+      key,
+      className: 'wallpaper-swatch',
+      style: {backgroundColor: backgroundColor},
```
paulrouget: `{backgroundColor}`

paulrouget:
You can land this.

src/browser/dashboard/actions.js
gordonbrander: Was trying to imitate `webviewer/actions.js` which uses short verbs for data model actions. I can change it though. While I'm at it, `createDashboard` is a better name.

src/browser/actions.js
gordonbrander: Would rather not. Purpose of function is to be able to add/change data model fields in a central location. Analogous to `webviewer/actions.js` `open({})` function.

src/browser/browser.js
gordonbrander: I thought it would be better not to pass cursor unless I need write access.

css/dashboard.css
gordonbrander: `flex-flow` just combines `flex-direction` and `flex-wrap` https://css-tricks.com/snippets/css/a-guide-to-flexbox/

paulrouget: Sure. But let's keep flex box simple and name what we use.

Issue #209, opened 2015-03-10
Determine update method


gordonbrander:
- We should have instant startup
- We should be able to update the UI on startup
  - Push updates?
- Caching approach... how do we mark interface dirty?

cc @paulrouget @Gozala

paulrouget:
@fabricedesre knows better, and he's working on that.

gordonbrander:
@gordonbrander need to design a notification style so we can tell user "update available".

fabricedesre:
So we have two pieces:
- updates of the runtime. This has landed, and updates are silently downloaded applied on restart. We could trigger a notification when updates are available. I didn't add that because there was no UI ready.

- updates of the front-end.  Currently this works by hitting F5.
We can do something like what Gordon describes (except the "don't reload iframes"), likely with a less aggressive polling. What I don't like here is that we need to maintain this "version file", and that's as error prone as an appcache manifest ;). If the version file can be autogenerated when pushing to gh-pages that would be nice.

A longer term option would be to use the push notifications api, and ensure that every time we push to gh-pages, me also trigger a push notification and use it to force refresh.

Overall, I wonder why just relying on http caching semantics is not good enough. This is what web sites are using... and we don't use any special caching rules for browser.html, so I would expect us to get the right version on every restart. Long lived session still need a solution though.

gordonbrander:
+1 on using HTTP caching if possible.

Regarding front-end, silent-update-on-restart is pretty good, but we would like to update in-app as a demonstration of what the technology can do.

When you do have to push an update to users, this type of mechanism is also a big improvement over the "we have updates -> restart" flow that we often force on users now. We could limit restarts to rendering engine updates, and even then only do it silently.

gordonbrander:
Closed by https://github.com/mozilla/browser.html/commit/fa1d697dc59e61e0046b5684c72c7fc95d984143

PR #243, opened 2015-03-13
Update mechanism

paulrouget:
Addressing issue #209

This is the first part. It's very basic, but good enough for a demo:

Github event API end point is pulled every 60 seconds, or longer, depending on what the server answers. We check if there's any new push to `gh-pages`, if there is, we show a banner with a `location.reload(true)` button. That's it.

A better strategy is necessary, but we're not there yet.

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -42,6 +42,11 @@
 
 /* Let's use flexbox everywhere */
 
+body {
+  height: 100vh;
+  overflow: hidden;
+}
+
 main {
   height: 100vh;
   width: 100vw;
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -41,7 +41,8 @@ define((require, exports, module) => {
     onDocumentBlur: Event('blur', getOwnerWindow),
     onDocumentKeyDown: Event('keydown', getOwnerWindow),
     onDocumentKeyUp: Event('keyup', getOwnerWindow),
-    onDocumentUnload: Event('unload', getOwnerWindow)
+    onDocumentUnload: Event('unload', getOwnerWindow),
+    onAppUpdateAvailable: Event('app-update-available', getOwnerWindow)
   });
 
   const onNavigation = KeyBindings({
@@ -177,7 +178,8 @@ define((require, exports, module) => {
 
     const suggestionsCursor = immutableState.cursor('suggestions');
 
-    return Main({
+    return DOM.div({
+    }, [Main({
       windowTitle: title(selectedWebViewerCursor),
       scrollGrab: true,
       className: ClassSet({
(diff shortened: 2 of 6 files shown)


css/theme.css
```diff
@@ -42,6 +42,11 @@
 
 /* Let's use flexbox everywhere */
 
+body {
+  height: 100vh;
+  overflow: hidden;
+}
```
Gozala: Nit: I have no idea why that's necessary, maybe you could add some comment explaining.

src/browser/browser.js
```diff
@@ -255,8 +258,22 @@ define((require, exports, module) => {
         onClose: item => webViewersCursor.update(closeTab(item)),
         onOpen: item => webViewersCursor.update(addTab(item))
       })
-    ]);
-  });
+    ]),
+    DOM.div({
```
Gozala: Is there any reason why we would not add these nodes only when update is available, instead of hiding it when it's not ?

src/browser/github.js
```diff
@@ -0,0 +1,69 @@
...
+  let interval = MIN_INTERVAL;
+  let pulling = false;
+
+  const pull = () => {
+    pulling = true;
+    let headers = {};
+    if (etag) {
+      headers = { "If-None-Match": etag }
+    }
+    fetch(API, { headers }).then(response => {
```
Gozala: Oh I wish I knew we had a `fetch` API available!

src/browser/browser.js
paulrouget: for the animation.

src/browser/index.js
```diff
@@ -6,12 +6,15 @@ define((require, exports, module) => {
 
   'use strict';
 
+  require('./github'); // Pull updates from github
```
Gozala: I would much rather prefer if this exported something along this lines:

```js
const update = () => {
  window.dispatchEvent(new CustomEvent("app-update-available"));
  require('./github').updateAvailable().then(update);
}

require('./github').updateAvailable().then(update);
```

So there would be a way to start / abort polling if necessary.

src/browser/index.js
```diff
@@ -6,12 +6,15 @@ define((require, exports, module) => {
...
+
   const {render} = require('./core');
   const {Browser} = require('./browser');
   const {readSession, resetSession} = require('./actions');
 
   // See issue #218
-  // render(Browser, readSession() || resetSession(), document.body);
-  render(Browser, resetSession(), document.body);
+  let session = readSession() || resetSession();
+  session = session.set('appUpdateAvailable', false);
```
Gozala: [`writeSession`](https://github.com/mozilla/browser.html/blob/master/src/browser/actions.js#L105-L107) should reset fields that don't need to be stored instead.

src/browser/browser.js
Gozala: > for the animation.

Can you add comment somewhere then so it's clear. Also in long term we should find a better way to handle that not just have everything in DOM at all times. I imagine react users have figured that out, I'll look into that.

Gozala:
@paulrouget I've added some comments and suggestion mainly to make things little more clear for a reader like me. That being said, I don't find anything to be important enough to block landing, so I'll leave it up to you to decide if you’d rather address some of these & then land or land as is (we can always clean some things up later).

src/browser/browser.js
paulrouget: Oh we can absolutely create the element and animate once mounted, but we can do that later.

paulrouget:
Merged with comments addressed fa1d697dc59e61e0046b5684c72c7fc95d984143.

Issue #196, opened 2015-03-06
Implement interim tab model


gordonbrander:
This is an in-between tab model... an evolutionary state between standard tabs and the full tab model as originally designed https://github.com/mozilla/browser.html/issues/112. We're implementing it because the full tab model requires some platform changes.

Basically, we use ordinary tabs. We don't open new tabs for different domains. But we move the most recently accessed tab to the front, making the tab strip order-by-last-use.

- Introduce the concept of "committing" to a tab. Clicking on a tab will select and activate it, but the tab will not be re-ordered until tab shade is hidden. This re-ordering after mousing out is called "committing"
- When closing, always activate the tab to the left
  - Tab to the left should always be the previous tab
  - (excluding tabs you clicked on but didn't commit to)
- New tabs
  - When opening, always open tab at the front
  - New tabs open as an "unfilled" tab
  - If not filled, unfilled tabs disappear when the tabstrip is hidden.
  - There should only be one unfilled tab in existence at a time. It's a temporary volatile state.

gordonbrander:
Note that the designed order is RTL, with newest at right. This matches the directionality of the back button `<`.

However, it is opposite of the OSX switcher and the LTR direction that `ctl tab` and `cmd shift ]` have in apps like Firefox and Sublime text. There are advantages to either direction. We'll test both. You can change the direction by getting rid of `* -1` here  https://github.com/mozilla/browser.html/pull/195/files#diff-303e3b2dd6791ebc293944dfa839d3a7R45

gordonbrander:
Closed by https://github.com/mozilla/browser.html/pull/195.

Given that https://github.com/mozilla/browser.html/issues/154 actually includes an interaction that is related to the tab model, I've moved it into this week's milestone.

Issue #112, opened 2015-02-19
Full tab model


gordonbrander:
Implement full tab model as designed.

In this model each domain is a tab. In fact, tabs aren't so much tabs as "active sites", in the same way that the task manager on iOS is your "active apps".

- New tabs are automatically opened for each root domain.
  - Clicking a link will either open a page in-place, or, if domain is different, it will open a new tab.
  - Exception: search pages belong to the tab who's page they open. They do not have a separate tab.
- You can open multiple tabs for the same domain the same way you do today: right click -> open in new tab, or cmd+click.
- Each tab has its own history stack (same as today). You can navigate it using controls in location bar.
- Tabs are ordered by last access from right to left, with the most recent at the far right.
- Hitting back on a tab that has no more "back" will transition leftward to the tab before it.
- After tabs fill the screen, old ones fall off. However they can be quickly brought back via tab search or dashboard.

Design rational
---------------------

Hypothesis: tabs today do many things, all poorly:

- Multitasking
- A place to dock a background task (music, email)
- Bookmark "limbo"

As a result:

- Tabs have to be organized by hand
- Tab explosion
- When one window fills up with tabs another has to be opened up. It's a windowing system that relies on another windowing system to function.
- Tabs are a windowing system. The browser is an OS. We need a richer interaction model than X-axis to be able to create a fully immersive browser that can disrupt use of the host OS.

Try:

- Fixing the multitasking case by focusing on the goal: quickly switching between 2-8ish of your most recently used sites.
- Later: supplement  with a tab dashboard that represents your full browsing history for the day. This should solve the "bookmark limbo" case.
- cmd+t new tab. cmd+shift+t switch tabs.

Why:

- Alt+tab, iOS multitasker and WebOS cards are all very successful multitasking  models because they self-organize. Both use the same interaction model:  your recently used active apps are always at the front.
- People think in domains. Let's organize pages by domain.
  - An app is a container for a branded experience.
  - A TV channel is a container for a branded experience.
  - Likewise a domain (facebook.com) is a container for a branded experience.
  - Analogy: dashboard is your inventory, tabs are your loadout.

Desired outcomes / questions we want to answer:

- How does task switching and clustering by domain feel during actual browsing?
- Be able to use the browser in a limited fashion.
- We will try another approach if: Gordon and Paul both think it's terrible.

Gozala:
[...]

PR #252, opened 2015-03-14
Implement new tab as a dashboard interface

Gozala:


Gozala:
@gordonbrander can you please take a look at this ?

P.S.: I think it this new tab idea does make a huge difference!

gordonbrander:
@Gozala it works!

It does make a big difference, even without a special style for the dashboard tab. Everything about spacial model feels clearer now. This is great news. I was unsure if the hybrid tab model #196 would feel right until we got to full tab model #112.

gordonbrander:
BTW, had to clear my session history for this to work. We should increment session history version.

If you don't clear session history, you get the following bug:

What should happen:
Website tile opens to right of dashboard tab on enter. Dashboard tab always at front.
Hitting `cmd t` opens dashboard tab, focuses location field

What happens:
Website tile opens __inside of__ of dashboard tab.
A new "new tab" is created _to the right_ of new website tile.
Hitting `cmd t` will focus current website tile, rather than opening dashboard.

src/browser/deck/actions.js
```diff
@@ -90,83 +96,136 @@ define((require, exports, module) => {
...
+
+  // Returns items sorted by their weight.
+  const arrange = items => items.sortBy(weightOf);
+
+  const center = (from, to) => (from + to) / 2;
+
+  // Returns given `item` with modified `weight` such that it's
+  // weight will fall in the center of `leading` & `following`
+  // item weights.
+  const asBetweenOf = (item, leading, following) => {
```
gordonbrander: Idea: `weightBetween(item, leading, following)` or `setWeightBetween(item, leading, following)` would be clearer naming scheme.

src/browser/deck/actions.js
```diff
@@ -214,24 +273,19 @@ define((require, exports, module) => {
...
-      index >= items.count() - 1 ? items.push(asLast(items, item)) :
-      items.push(asNth(items, index, item));
-
-    const update = compose(include,
-                           isSelected(item) ? deselect : identity,
-                           isActive(item) ? deactivate : identity);
-
-    return update(items);
+  const include = (items, item) => {
+    const edit = compose(items => items.push(item),
```
gordonbrander: Possible bug? `Array.prototype.push` returns [new length property of the object upon which the method was called](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push). Since you are composing, I would expect you need to return `item`.

src/browser/page-switch.js
```diff
@@ -43,6 +43,7 @@ define((require, exports, moudle) => {
       ]),
       DOM.div({
         key: 'close-button',
+        hidden: webViewerCursor.get('isPinned'),
```
gordonbrander: Idea: perhaps we should not render if item is not pinned? No big deal though.

gordonbrander:
This is the PR for https://github.com/mozilla/browser.html/issues/154

src/browser/deck/actions.js
Gozala: It's not an array but immutable list, all methods return new lists in this case with item added

gordonbrander: Cool. Might be worth adding a note about this in a comment, but whatever you think is best.

src/browser/page-switch.js
Gozala: > Idea: perhaps we should not render if item is pinned? No big deal though.

Yeah good call.

gordonbrander:
R+

PR #256, opened 2015-03-14
Change default theme

gordonbrander:
- Change default theme to something vanilla.
- Fix theme code so it will fall back to sensible defaults.


gordonbrander:
@Gozala R+'d it.

PR #258, opened 2015-03-14
Add new tab style

gordonbrander:
Initial new tab style. It's not perfect, but it's better than what we had.

We unified the concept of "dashboard", "new tab", "add tab" into a single model. I prefer the "+" analogy to something like a home icon because it is the most immediately aligned with a new user's interest ("I want a new tab"), with the discovery of the dashboard being a happy result.

### DIFF
--- a/css/tabstrip.css
+++ b/css/tabstrip.css
@@ -103,3 +103,32 @@
   transition: opacity 150ms 250ms ease;
   pointer-events: auto;
 }
+
+.tab-dashboard .tab-thumbnail {
+  background: transparent;
+  border: 1px solid rgba(0,0,0,0.1);
+}
+
+.tab-dashboard .tab-thumbnail::after {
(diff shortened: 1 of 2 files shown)


gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/254

css/tabstrip.css
```diff
@@ -103,3 +103,32 @@
...
+  background: transparent;
+  border: 1px solid rgba(0,0,0,0.1);
+}
+
+.tab-dashboard .tab-thumbnail::after {
+  color: rgba(0,0,0,0.2);
+  content: "\f067";
+  cursor: default;
+  display: block;
+  font: 24px/62.5px FontAwesome;
```
paulrouget: Ok, but in the future, I'd prefer if we could avoid shorthands for fonts too.

paulrouget:
r+

PR #259, opened 2015-03-15
Implement a better debugging support.

Gozala:
That way I can for example I could define function that would log subpart of state I'm interested on every change, or even have conditionals to log only when entering a broken state etc..


Gozala:
As an example I use this now to debug issue of colliding weights

```js
window.debug = (state) => console.log(state.toJSON().webViewers.map(x => x.weight).sort().join('\n'))
```

paulrouget:
r+

PR #267, opened 2015-03-15
Fix for new tab as dashboard feature 

### LINKED ISSUE #251: I think removing a tab may lead to a bug
We've run into [a bug](https://github.com/omniscientjs/omniscient/issues/89) in the cursor / omniscient implementation that was caused by re-ordering items in the list. Problem is that cursors are basically a proxy to a `path` under the `target` object that implements same API as structure under that path. So when items get reordered old `x` cursor will have same value but a different path, from new `x` cursor. Then if component is passed that `x` it won't run but return from cache since values matched. Which means that component will point to old `x`. On user event update to `x` will be send to the renderer but since that `x` will have an old `path` renderer will end up updating a wrong thing and causing an error.

To workaround that I end up ditching reordering of items in favour of encoding `order` into item itself, which resolved the issue.

Although now I realised that when tab is closed items will still reorder in the list which may lead to same issues as described above. Not sure why we have not seen it yet but I fear it's there.

Gozala:
Initially tab / web-viewer deck items were represented in form of [Immutable.List](http://facebook.github.io/immutable-js/docs/#/List) which worked fine at first. Later when we started arranging items by last activation we discovered [bug in the cursor implementation](https://github.com/omniscientjs/omniscient/issues/89). In order to work around that last implementation no longer re-ordered items, instead they were always eded to the tail and `weight` property on each item was set to encode the position. Unfortunately logic behind the `weight` did not take into account that our ordering always puts items to the left there for it's inevitable to hit the issue where all items will end up having same `weight`.

In this pull request instead of workarounding a bug in cursor implementation I avoid using cursors all together (as suggested here https://github.com/mozilla/browser.html/issues/250#issuecomment-80030040) the tab / web-viewer decks and switch back to the list re-arranging approach. There for it fixes issue in the last implementation that depended on a false `weight` logic.

Please note that even though I said I avoid using cursors, I don't mean that they are fully ripped out as that would require major code changes, instead it just wraps cursors at the dec level with a function, that way updates will cause handler function to be hot-swapped without causing re-render while when invoked they will actually refer to a up to date cursor since any change to the items will re-trigger deck and there for generate new functions.

This also should fix #251 as that one is caused by the same cursor bug.


paulrouget:
r+

Issue #264, opened 2015-03-15
Session data bug


PalmerAL:
With #252, it's apparently possible to end up in a state where the tab strip is never shown and it is impossible to create new tabs (happened twice so far). Only way to fix is to reset with shift-command-delete. Haven't found reliable steps to reproduce yet.

Gozala:
@paulrouget @PalmerAL if you run into this again could you please paste state dump it can give a clue what's going on. You can do it by running following in the console connected with a webide

```js
JSON.stringify(state, null, 2)
```

PalmerAL:
What's a good way to launch this then? In order to connect to webide, you have to (I think) run off of local host, but if you launch that from console, keystrokes in the window actually end up in console, making it impossible to use. Is there a way to run the command in the readme that lets keystrokes be applied to browser.html and not the terminal, or is there a way to do this with the github pages as the start manifest?

paulrouget:
No need to run from local host. Devtools are started no matter what. Just run the app, then open WebIDE from a regular Firefox, and connect to localhost:6000

Gozala:
I think the issue described maybe what the #268 is fixing. Although I'm not certain.

@PalmerAL 

>  where the tab strip is never shown and it is impossible to create new tabs (happened twice so far).

If that is the same issue I think what you may have bing observing is that address bar was in edit mode, and in that case hovering it does not bring up tabstrip (@gordonbrander is that desired BTW) although switching through should still be possible with keybindings. 

Also what do you meant by "impossible to create a new tab" ? Does `cmd + t` did nothing ? If that's a same bug as #268 I think you will get a new tab but you won't be able to type much as the selection bug will select whatever you type and you'll basically end up changing one character.


Does this sounds like what you have observed ?

Gozala:
I've also created #269 to add a poor mans view-source or rather state-source so state dump could be obtained without having to open devtools.

PalmerAL:
when this happens, cmd+t doesn't work. It just selects the URL (like cmd+l normally would).

PalmerAL:
I'm thinking this may actually be caused by #271, as you can't create new tabs when the dashboard new tab is gone.

PalmerAL:
However with that issue you can still see the tabstrip. Probably, this is #271 and #268 combined together.

paulrouget:
Just happened to me: http://jsoneditoronline.org/?id=a01913a5da45e3c82782655f6b6145c6

paulrouget:
[...]

PR #268, opened 2015-03-15
Selection state should be updated both ways.

Gozala:
Sometimes you can end up in broken state where escaping the address bar is impossible. I believe it could be reproduced by pressing `cmd l` while in the location bar. I think this is also what issue #264 is about, but I maybe wrong.

Either way current problem is that keyboard shortcut will store `selection` state into a state that later is used to update selection in the input. The problem was that changes to the selection were never reflected in the state there for any re-render would select all of the text in the input over and over.

This patch cleans up a little how selection was implemented (it was hacked together without much of a thought) and wires both ends so that changes to selection are reflected in the state and changes in the state are reflected in the UI.

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -108,7 +108,6 @@ define((require, exports, module) => {
   exports.readInputURL = readInputURL;
   exports.focus = focusable => focusable.set('isFocused', true);
   exports.blur = focusable => focusable.set('isFocused', false);
-  exports.select = editable => editable.set('selection', {all: true});
   exports.showTabStrip = tabStripCursor =>
     tabStripCursor.set('isActive', true);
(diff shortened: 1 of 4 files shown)


paulrouget:
r+

paulrouget:
@Gozala ,

I understand why we do this, but the fact we have to record changes from the page (`.onSelect`) bugs me a bit. I ran into the same problem with the scroll position (needing to listen to `onScroll()`). Without React, we would just do `.select()` or `.scrollTop = 0`, but with React, we need to use the state, and update it, otherwise, we would keep calling `select()/scrollTop`.

For these kind of actions, do we really have to use the state? Can't we just call `.select()` directly from the event?

PR #275, opened 2015-03-15
Prevent closing dashboard via cmd+w

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/271

@Gozala I can also put together a follow-up that factors out `remove(items, p)` into `findAndRemove(items, p)` and `remove(items, item)`. That would be marginally more efficient since we wouldn't have to `find` twice.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -107,8 +107,7 @@ define((require, exports, module) => {
   // This avoids code branching down the pipe that otherwise will
   // need to deal with 0 viewer & no active viewer case.
   const close = p => items =>
-    items.count() > 1 ? remove(items, p) :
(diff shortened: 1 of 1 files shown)


Gozala:
r+

Gozala:
> I can also put together a follow-up that factors out remove(items, p) into findAndRemove(items, p) and remove(items, item). That would be marginally more efficient since we wouldn't have to find twice.

@gordonbrander I think I'd prefer to modify a keybinding from this:
https://github.com/mozilla/browser.html/blob/master/src/browser/browser.js#L149

To something like:

```js
{
  'accel w': edit(close(and(isActive, isntPinned))),
}
```

And make sure (modify if necessary) that `remove` handles case with no match properly.

Issue #266, opened 2015-03-15
Can't take screenshot unless iframe was ever displayed


Gozala:
Now that we have session restore back on I was trying to figure out why we don't get thumbnails when you refresh. It looks like `.getScreenshot()` call remains forever pending if iframe always had `display: none`. Note that if iframe was displaying event once this issue no longer occurs.

I came to above conclusion by doing following reloading `browser.html` that already had some pages open with tabs tiles generated out of screenshots. After reload I connected with wedIDE & used console to try to `.getScreenshot()` note that calls to `setVisible` and `setActive` had no effect on the outcome. Finally after removing `hidden` attribute and then putting it back `.getScreenshot()` did succeed.

@paulrouget is this a bug that we can solve or a limitation of the platform ? If later then we should probably stack iframes on top of each other and and hide inactive iframes only if they were loaded.

Gozala:
@paulrouget do we want to keep this one open ? Workaround in #277 works although I'd prefer if we could avoid such workarounds.

paulrouget:
Let's keep this open.

Gozala:
servo/servo#7083

Gozala:
https://bugzilla.mozilla.org/show_bug.cgi?id=1198503

Gozala:
Our target runtime is on Servo so no point in keeping gecko issues open.

PR #277, opened 2015-03-16
Fix tiles on refresh.

Gozala:
This is a workaround fix for #266

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -7,7 +7,7 @@ define((require, exports, module) => {
   'use strict';
 
   const url = require('./util/url');
-  const {fromJS} = require('immutable');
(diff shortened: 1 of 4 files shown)


src/browser/actions.js
```diff
@@ -96,9 +96,28 @@ define((require, exports, module) => {
...
+        title: null,
+        backgroundColor: null,
+        foregroundColor: null,
+        isDark: false,
+        securityState: 'insecure',
+        securityExtendedValidation: false,
+        canGoBack: false,
+        canGoForward: false
+      }))).
+      toJSON();
```
paulrouget: Instead of resetting things, we should only store things we want to keep across session (only locations afaict), and merge on restore.

PR #283, opened 2015-03-16
Update dependencies.

Gozala:


paulrouget:
@Gozala feel free to merge, but please also update gh-pages

Gozala:
Discussed with @paulrouget and we agreed to not push to gh-pages just yet to make sure there were no regressions caused by it.

PR #285, opened 2015-03-16
run npm install before python

yukinying:
The http server require node_module to be presented, otherwise index.html would not load.


Gozala:
Seem good to me, @yukinying would you mind updating your pull as it seems to conflict with current master as is.

paulrouget:
README now includes npm install.

yukinying:
Thanks! I was out for a while and not able to update the patch.

PR #290, opened 2015-03-18
deploy

### LINKED ISSUE #262: Sync gh-pages with travis
I think Travis can automatically merge master into gh-pages, and run `npm install`.

I know nothing about these things. Can anyone help?

paulrouget:
Auto deploy to gh-pages with Travis (fix #262).
This add `node_modules` and a `HEAD` file that includes the last commit hash. It will be used to know what is the current revision from browser.html.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,11 @@
 language: node_js
 node_js:
-  - "0.10"
+  - '0.10'
+script: npm run-script deploy
+branches:
+  only:
+    - master
+env:
+  global:
+  - GH_REF: github.com/mozilla/browser.html.git
+  - secure: HVI/ynSYdnb6GGm6uBP/+9G0ZMn/cW95iiyLoz6C2uNS62Ec7F0u3tAkNaIaOXXU18bJ4i3Y3w0Vcw7FRYOXR60MI3J9SRtkuVjhhetE/S8UI1iEgfW/1/+e/poQzRXMut/1o5xS4V8t1ZDSwet5BtWIDdy9K6GUES5b3hSdQjk=
--- a/css/updatebanner.css
+++ b/css/updatebanner.css
@@ -3,18 +3,19 @@
   position: absolute;
   justify-content: space-between;
   bottom: 10px;
-  width: 300px;
-  left: calc(50vw - 150px);
+  width: 400px;
+  left: calc(50vw - 200px);
   background-color: rgba(36,60,83,0.95);
   border-radius: 4px;
   padding: 8px 8px 8px 20px;
   color: white;
   align-items: center;
-  transition: transform 500ms ease-in;
+  transition: opacity 500ms ease-in;
 }
 
 .appupdatebanner:not(.active) {
-  transform: translateY(200px);
+  opacity: 0;
+  pointer-events: none;
 }
 
 .appupdatebanner > span {
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -e
+set -v
+
+git config --global user.email "travis@browser.html"
+git config --global user.name "travis"
+git rev-parse HEAD > HEAD
+git checkout --orphan gh-pages
+rm .gitignore
+git add .
+git commit -m "Deployed to Github Pages"
+git push --force --quiet "https://${GH_TOKEN}@${GH_REF}" gh-pages > /dev/null 2>&1
+git status
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -46,7 +46,8 @@ define((require, exports, module) => {
     onDocumentKeyDown: Event('keydown', getOwnerWindow),
     onDocumentKeyUp: Event('keyup', getOwnerWindow),
     onDocumentUnload: Event('unload', getOwnerWindow),
-    onAppUpdateAvailable: Event('app-update-available', getOwnerWindow)
+    onAppUpdateAvailable: Event('app-update-available', getOwnerWindow),
+    onRuntimeUpdateAvailable: Event('runtime-update-available', getOwnerWindow)
   });
 
   const onNavigation = KeyBindings({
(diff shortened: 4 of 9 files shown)


paulrouget:
So, client side, when we get a `pushEvent` from `https://api.github.com/repos/mozilla/browser.html/events` (only for the `gh-pages` branch), we could then compare `/HEAD` to `https://api.github.com/repos/mozilla/browser.html/git/refs/heads/master`. If they defer, we clear the cache and reload the page.

Does that make sense?

Maybe we could avoid the request to `refs/heads/master` by looking at the payload of `mozilla/browser.html/events` and look at the pushes that happened in master.

We can't just look at the pushes to `master` because it takes time for travis to push to `gh-pages`. And we can't use the HEAD of `gh-pages` because we actually change it by adding `node_modules` and the `HEAD` files.

/cc @Gozala @donaldpipowitch

paulrouget:
This is how it works:

1. pushing to master will trigger a travis job, where we create a new file, called `HEAD`, that includes the hash of the HEAD from master, then push it to `gh-pages`. So `gh-pages` is always in sync with master, with 2 extra things: the `HEAD` file and `node_modules` (automatically installed by travis).
2. client side, everything is cached, including the HEAD file. Github provides an API to request the content of a repo (not cached). So it's easy to compare the local HEAD file to the remote HEAD file. If they differ, there's an update.

css/updatebanner.css
```diff
@@ -3,18 +3,19 @@
...
   color: white;
   align-items: center;
-  transition: transform 500ms ease-in;
+  transition: opacity 500ms ease-in;
 }
 
 .appupdatebanner:not(.active) {
-  transform: translateY(200px);
+  opacity: 0;
+  pointer-events: none;
```
paulrouget: removing `transform` as it broke scrolling. I'll fix that later.

.travis.yml
```diff
@@ -1,3 +1,11 @@
...
-  - "0.10"
+  - '0.10'
+script: npm run-script deploy
+branches:
+  only:
+    - master
+env:
+  global:
+  - GH_REF: github.com/mozilla/browser.html.git
+  - secure: HVI/ynSYdnb6GGm6uBP/+9G0ZMn/cW95iiyLoz6C2uNS62Ec7F0u3tAkNaIaOXXU18bJ4i3Y3w0Vcw7FRYOXR60MI3J9SRtkuVjhhetE/S8UI1iEgfW/1/+e/poQzRXMut/1o5xS4V8t1ZDSwet5BtWIDdy9K6GUES5b3hSdQjk=
```
paulrouget: This key is an encrypted (with travis public key) github token.

paulrouget:
@gordonbrander Do you think you can review that?

css/updatebanner.css
```diff
@@ -3,18 +3,19 @@
   position: absolute;
   justify-content: space-between;
   bottom: 10px;
-  width: 300px;
-  left: calc(50vw - 150px);
+  width: 400px;
```
gordonbrander: Just pointing out width change in case it was a mistake. Otherwise, ok.

deploy.sh
```diff
@@ -0,0 +1,14 @@
...
+set -v
+
+git config --global user.email "travis@browser.html"
+git config --global user.name "travis"
+git rev-parse HEAD > HEAD
+git checkout --orphan gh-pages
+rm .gitignore
+git add .
+git commit -m "Deployed to Github Pages"
+git push --force --quiet "https://${GH_TOKEN}@${GH_REF}" gh-pages > /dev/null 2>&1
```
gordonbrander: Why force? Force is scary.

src/browser/browser.js
```diff
@@ -297,15 +299,27 @@ define((require, exports, module) => {
...
+          }
+          if (immutableState.get('runtimeUpdateAvailable') && !immutableState.get('runtimeUpdateAvailable')) {
+            // FIXME: Not supported yet
+            sendEventToChrome('restart')
+          }
+          if (!immutableState.get('runtimeUpdateAvailable') && immutableState.get('runtimeUpdateAvailable')) {
+            sendEventToChrome('clear-cache-and-reload')
+          }
+        }
+      }, 'Apply' + (immutableState.get('runtimeUpdateAvailable') ? ' (restart required)' : ''))
```
gordonbrander: This works for now. I would chat about the UX around updates. I want users to feel like they never cost anything.

One strategy could be to only do runtime updates when you quit the browser. That way you never have to restart. This would require tagging app updates with required runtimes.

src/browser/github.js
```diff
@@ -7,67 +7,73 @@ define((require, exports, module) => {
...
+      } else {
+        reject(`Can\'t reach HEAD: ${response.statusText}`);
+      }
+    }).catch(reject);
+  });
+
+  const appUpdateAvailable = new Promise(pull);
 
-  exports.appUpdateAvailable = new Promise(pull);
+  appUpdateAvailable.then(() => clearTimeout(timeout));
```
gordonbrander: Will this mean if app gets one update, it will stop polling until you restart runtime?

src/browser/index.js
```diff
@@ -9,10 +9,10 @@ define((require, exports, module) => {
   const {render} = require('./core');
   const {Browser} = require('./browser');
   const {readSession, resetSession} = require('./actions');
-  const {appUpdateAvailable} = require('./github'); // Pull updates from github
 
   render(Browser, readSession() || resetSession(), document.body);
 
+  const {appUpdateAvailable} = require('./github'); // Pull updates from github
```
gordonbrander: Better to keep all the imports together. This file didn't need to change.

src/browser/browser.js
Gozala: Note that some updates may depend on runtime updates so that would need to be expressed some how as well.

deploy.sh
paulrouget: Because we clear the branch each time. We don't pull the existing `gh-pages` branch (there's no need), we only clone master.

src/browser/browser.js
paulrouget: We'll solve that later.

src/browser/github.js
paulrouget: Yes. There's no point of "knowing" an update is available once one is available.

PR #294, opened 2015-03-19
Basic support for about:setting

paulrouget:
This comes with 2 commits:

* Support for `about:*` urls. `about:foobar` will load `http://…/src/about/foobar/index.html` in a privileged tab, which has the same privileges as browser.html.

* `about:settings` page. Just dump the settings. Very basic, but will be useful to check versions & co.

### DIFF
--- a/src/about/settings/index.js
+++ b/src/about/settings/index.js
@@ -1,0 +1,53 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+define((require, exports, module) => {
+
+  'use strict';
+
+  const Component = require('omniscient');
+  const {render} = require('browser/core');
+  const {DOM} = require('react');
+
+  const {Map} = require('immutable');
+  const Table = Component('Table', state => {
+
+    let rows = [];
+    for (let [name,value] of state) {
+      rows.push(DOM.div({
+        key: name,
+        className: 'row'
+      }, [
+        DOM.span({
+          key: 'name.' + name,
+          className: 'name',
+          title: name
+        }, name),
+        DOM.span({
+          key: 'value.' + name,
+          className: 'value',
+          title: value
+        }, value)
+      ]));
+    }
+
+    return DOM.div({
+      key: 'table',
+      className: 'table'
+    }, rows);
+
+  });
+
+  const table = render(Table, new Map(), document.body);
+
+  navigator.mozSettings
+           .createLock()
+           .get('*')
+           .then(r => table.step(state => state.merge(r)));
+
+  navigator.mozSettings.onsettingchange = e => {
+    table.step(state => state.set(e.settingName, e.settingValue));
+  }
+
+});
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -18,10 +18,14 @@ define((require, exports, module) => {
   const makeSearchURL = input =>
     `https://duckduckgo.com/?q=${encodeURIComponent(input)}`;
 
+  const makeAboutURL = input =>
+    url.getBaseURI() + 'src/about/' + input.replace('about:', '') + '/index.html';
+
   const sendEventToChrome = type => dispatchEvent(new CustomEvent('mozContentEvent',
     { bubbles: true, cancelable: false, detail: { type }}))
 
   const readInputURL = input =>
+    url.isAboutURL(input) ? makeAboutURL(input) :
     url.isNotURL(input) ? makeSearchURL(input) :
     !url.hasScheme(input) ? `http://${input}` :
     input;
--- a/src/browser/util/url.js
+++ b/src/browser/util/url.js
@@ -37,6 +37,12 @@ define(function() {
       return this.a.origin;
     },
 
+    getBaseURI: function urlHelper_getBaseURI() {
+      // http://example.com/foo/index.html -> http://example.com/foo/
+      // http://example.com/foo/ -> http://example.com/foo/
+      return location.href.replace(/\/[^/]*$/, '/');
+    },
+
     getHostname: function urlHelper_getHostname(url) {
       this.a = this.a || document.createElement('a');
       this.a.href = url;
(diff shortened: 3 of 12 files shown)


src/about/settings/index.js
```diff
@@ -0,0 +1,61 @@
...
+    return DOM.div({
+      key: 'table',
+      className: 'table'
+    }, rows);
+
+  });
+
+  let state;
+
+  const update = () => render(Settings, state, document.body);
```
Gozala: Idea was that one would call `render` to spawn up a render loop, this works but seems little odd. I would advice doing something along these lines instead:

```js
const settings = render(Settings, state, document.body)
```

Maybe `settings` isn't a great name, I'll let you come up with a good one. Other one that comes to my mind is `app`.

src/about/settings/index.js
```diff
@@ -0,0 +1,61 @@
...
+
+  navigator.mozSettings
+           .createLock()
+           .get('*')
+           .then(r => {
```
Gozala: Then from here you could do:

```js
settings.step(_ => fromJS(r))
```

src/about/settings/index.js
```diff
@@ -0,0 +1,61 @@
...
+             state = fromJS(r);
+             update();
+           });
+
+  navigator.mozSettings.onsettingchange = e => {
```
Gozala: And this will become a lot cleaner as well:

```js
settings.step(state => state.set(e.settingName, e.settingValue))
```

src/browser/web-viewer.js
```diff
@@ -69,7 +69,13 @@ define((require, exports, module) => {
       onAuthentificate: WebViewer.onAuthentificate(edit),
       onScrollAreaChange: WebViewer.onScrollAreaChange(edit),
       onLoadProgressChange: WebViewer.onLoadProgressChange(edit)
-    })
+    }
+
+    if (url.isPrivileged(state.get('uri'))) {
+      dom.mozApp = url.getManifestURL();
+    }
```
Gozala: Could not you just inline this instead like:

```js
IFrame({
...
  mozApp: url.isPrivileged(state.get('uri')) ? url.getManifestURL() : null
```
})
```

File: src/browser/web-viewer.js [hunk 5]
```diff
@@ -206,7 +212,9 @@ define((require, exports, module) => {
   WebViewer.Deck = Component('WebViewerDeck', (options, {onOpen, onClose, edit}) => {
     const {items, In} = options;
     return DOM.div(options, items.sortBy(id).map(item => WebViewer({
-      key: item.get('id'),
+      // Change key depending on privileges,
+      // forcing a re-creation of the iframe if privilege level change
+      key: item.get('id') + (url.isPrivileged(item.get('uri')) ? 'PRIVILEGED' : 'NOTPRIVILEGED'),
```
Gozala: This works, although I don't think forcing re-rendering as such is very clean & not easy to scale if number of properties will grow.

The real problem here is that there are non-standard attribute changing which at the DOM level has no effect. I think it would de best to fix it actually on the platform level. In the meantime I think it would be best to change our `Element` implementation to handle such cases better, in fact our [`BeforeAppendAttribute`](https://github.com/mozilla/browser.html/blob/master/src/browser/element.js#L89-L107) is kind of does that already so what I'd do is:

1. Change it's name be something more adequate like `RemountAttribute` or something like that.
2. Add a `write` method to it's implementation that would [call a super method](https://github.com/mozilla/browser.html/blob/master/src/browser/element.js#L128-L136) and then re-inject node to force it to be picked up:

```js
BeforeAppendAttribute.prototype = {
    constructor: BeforeAppendAttribute,
    mount(node, value) {
      node.setAttribute(this.name, value);
    },
   write(node, present, past) {
     Attribute.prototype.write.call(node, present, past)
     if (present !== past) {
       node.parentNode.replaceChild(node, node)
     }
   }
  };
```

This way we'll force node re-creation for attributes that require that. That way user of component like `IFrame` does not need to know intimate details of `IFrame` implementation to make use of them.

Gozala:
I made a bunch of comments inline. Another thing that I wanted to point out is that I am not sure we need full fledged document for `about:settings` or other `about:` views. I think maybe about components would be enough that would avoid whole new configuration and copies of the same JS modules, although I do see one complication with it which is we should probably stop listening for a setting changes if `about:settings` isn't rendered. I think it won't be too hard to define something like `unmount` hook for that purpose though or alternatively owner component could deal with that by starting / stopping certain listener depending on weather child was rendered or not.
That being said I'm happy to do it in the next iteration or not at all if you there is a reason why having separate document is better.

src/browser/actions.js
```diff
@@ -18,10 +18,14 @@ define((require, exports, module) => {
   const makeSearchURL = input =>
     `https://duckduckgo.com/?q=${encodeURIComponent(input)}`;
 
+  const makeAboutURL = input =>
+    url.getBaseURI() + 'src/about/' + input.replace('about:','') + '/index.html'
```
fabricedesre: uber nit: input.replace('about:', '') + '/index.html';

src/browser/util/url.js
```diff
@@ -87,6 +93,19 @@ define(function() {
       return !this.urlValidate.validity.valid;
     },
 
+    isAboutURL: function(url) {
+      return url.startsWith('about:');
```
fabricedesre: Ideally should use window.URL() and check the protocol (https://developer.mozilla.org/en-US/docs/Web/API/URLUtils/protocol) since it's case insensitive.

src/browser/util/url.js
```diff
@@ -87,6 +93,19 @@ define(function() {
...
+
+    isPrivileged: function urlHelper_isPrivilegedURI(uri) {
+      // FIXME: not safe. White list?
+      return uri && uri.startsWith(this.getBaseURI() + 'src/about/');
+    },
+
+    getManifestURL: function urlHelper_getManifestURL() {
+      return this.getBaseURI() + 'manifest.webapp';
+    },
+
```
fabricedesre: here also, that would be cleaner to resolve 'src/about/' and 'manifest.webapp' against getBaseURI() using window.URL().

paulrouget: Not sure what you mean. window.URL() can't be used to resolve urls.

paulrouget:
@Gozala I'd prefer if we could keep `about:*` pages in their own tabs. I don't see any benefit to keep them in the main process.

src/browser/util/url.js
fabricedesre: sure it can: url = new URL(urlString, [baseURLstring])

paulrouget: Oh cool! Didn't know that. I'll update the code.

PR #301, opened 2015-04-01
use window.URL in url.js

paulrouget:
@fabricedesre can you take a look?

### DIFF
--- a/src/browser/util/url.js
+++ b/src/browser/util/url.js
@@ -10,87 +10,59 @@ define(function() {
 
   'use strict';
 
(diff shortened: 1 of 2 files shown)


src/browser/util/url.js
```diff
@@ -10,87 +10,64 @@ define(function() {
...
+      const case4Reg = /^\w+\:\/*$/;
+      if (str == 'localhost') {
+        return false;
+      }
       if (case1Reg.test(str) || !case2Reg.test(str) || case4Reg.test(str)) {
         return true;
       }
       if (case3Reg.test(str)) {
         return false;
       }
```
fabricedesre: do we really need case3Reg at all? data: urls should be recognized properly by new URL()

PR #305, opened 2015-04-03
Use ecstatic instead of python SympleHTTPServer

Gozala:
Let's fully leverage `npm` + `node`  setup and kill dependency on python and SimpleHTTP.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,11 +12,13 @@
     "url": "http://github.com/mozilla/browser.html/issues/"
   },
   "scripts": {
-    "server": "python -m SimpleHTTPServer 8080",
-    "start": "/Applications/B2G.app/Contents/MacOS/graphene --start-manifest=http://localhost:8080/manifest.webapp",
+    "start-server": "ecstatic . --port 6060",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -12,11 +12,13 @@
     "url": "http://github.com/mozilla/browser.html/issues/"
   },
   "scripts": {
-    "server": "python -m SimpleHTTPServer 8080",
-    "start": "/Applications/B2G.app/Contents/MacOS/graphene --start-manifest=http://localhost:8080/manifest.webapp",
+    "start-server": "ecstatic . --port 6060",
```
gordonbrander: Uses global ecstatic I think. Anyway, this does not work for me.

Gozala: No npm creates `node_modules/.bin` folder and also includes it in path for `npm command` stuff so it's going to use the `ecstatic` that it installed and included in it's own path ;)

gordonbrander: ...Sent error message to you

gordonbrander:
r+

PR #309, opened 2015-04-07
group common files in a common folder

paulrouget:
Eventually, I'd like to get something like that:
```
.
└── src
    ├── common           // Components, modules, helpers, whatever
    ├── about
    │   └── settings     // Components specific to about:settings
    ├── browser          // Components specific to browser
    └── dashboard        // Components specific to dashboard
```

Here I move code from browser to common.


Gozala:
I am not sure why `lang` or `shims` ended up under `common` and why `dashboard` is it's own thing and not part of `browser`. That being said I don't have strong feelings either way about the layout.

paulrouget:
> I am not sure why lang or shims ended up under common

Right. I will move them directly into `src/`.

> and why dashboard is it's own thing and not part of browser

It is not.

paulrouget:
I see what you meant by dashboard having its own directory. It's not part of this PR. This will happen later because I want the dashboard to live in a tab (like `about:` pages). But that's for later.

PR #311, opened 2015-04-08
Illustration of how we could use records to model structured state

Gozala:
Right now this just updates awesome bar as it was simplest to update.


paulrouget:
I like that. I think we should build a schema for the state base on that.

PR #312, opened 2015-04-13
CSS to JS

paulrouget:


Gozala:
Makes sense to me, although I wonder if react has some built-in helpers so we don't need to use `Object.assign` everywhere.

src/browser/WindowControls.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const styleGreyButton = Object.assign({
+    backgroundColor: 'hsl(0, 0%, 86%)'
+  }, styleButton);
+
+  const WindowControls = Component(({isDocumentFocused, windowControls, theme}) => {
+
+    const styleMinButton = Object.assign({}, theme.windowMinButton, styleButton);
+    const styleMaxButton = Object.assign({}, theme.windowMaxButton, styleButton);
+    const styleCloseButton = Object.assign({}, theme.windowCloseButton, styleButton);
```
Gozala: I think it would be useful add an utility function to our toolchain similar to `classSet` for mixing styles, maybe something along these lines:

```js
const isObject = x => x && typeof(x) === "object"
const mix = (...styles) => Object.assign({}, ...styles.filter(isObject))

const styleCloseButton = mix(theme.windowCloseButton, styleButton)
```

PR #318, opened 2015-04-16
Reset readyState to fix repeated commands.

Gozala:
Currently there are bunch of issues caused by `readyState` field not being reset, for example if you do refresh twice second one is ignored that's because `readyState` value will be `refresh` instead of `loaded`. Similar issues happen with `goBack` `goForward`.


paulrouget:
r+

Issue #250, opened 2015-03-14
Proposal for some architectural changes


Gozala:
I decided to start this issues after some discussion with @paulrouget about architecture and code base. I think we (at least I) learned few things from the experience with a current code base & I would like to address some of the constraints & ambiguities. What will follow in the comment is more of a brain dump & likely some followup discussion.

Gozala:
The reason we use [immutable data structures](http://facebook.github.io/immutable-js/) for a several reasons:

1. To make it easy to reason about app state by avoiding out of sync situations, caused by mutations from one part of code that was not noticed by other part of the code.
2. Immutable data structures make it really easy to represent whole application state as a single data structure, and updates could be made deep down in the nested structure that would produce whole new structure as the root with common parts shared with previous. In other words making deep down updates very convenient.
3. By leveraging immutable data structures we take advantage of predefined & optimised [shouldComponentUpdate](http://facebook.github.io/react/docs/component-specs.html#updating-shouldcomponentupdate) to gain performance boost by not re-triggering components who's state has not changed, this is explained in further details by [PureRenderMixin](http://facebook.github.io/react/docs/pure-render-mixin.html)

Although since UI interaction need to cause state changes that somehow need to initiate re-render with another updated state, we employed [Cursors](https://github.com/facebook/immutable-js/tree/master/contrib/cursor) that is part of Immutable.js library that allows to present sections of the actual immutable data structure such that transformations of it will notify a [renderer](https://github.com/mozilla/browser.html/blob/master/src/browser/core.js) of a transformed state so it could re-render with updated state. Unfortunately use of `Cursor`s caused some ambiguity (for developers & the system):

1. It is not obvious when cursors should be used over plain data structures. Although [attempt](https://github.com/mozilla/browser.html/wiki/Design-F.A.Q.#q-how-do-i-cause-side-effect--re-render-) was made to make it more clear.
2. Cursors have a `read+write` semantics and when passed around it is not clear if they intended to be read from or to be written into.
3. Result of render purely depends on input meant for `read` anything `write` related does not need to re-render, but since cursors are `read+write` ones intended for `write` would cause re-renders as well. To avoid that we established [event handler passing pattern](https://github.com/mozilla/browser.html/wiki/Design-F.A.Q.#q-how-do-update-state-that-i-dont-need-to-render-) which makes things worse by having two different ways to update.
[...]

PR #319, opened 2015-04-16
Unite location and uri into single field.

Gozala:
This is related to discussion in #250 regarding a webview state simplification that drops `location` field in favor of `uri`.

This change basically moves hack from web view into iframe which is required to avoid changing a `iframe.src` if new value matches curretly loaded content `uri`.


paulrouget:
r+

PR #324, opened 2015-04-18
Refactor code according to a proposal.

Gozala:
I'm not totally sure I have not introduced any bugs here as I have not done proper testing but this is more or less what we agreed upon here https://github.com/mozilla/browser.html/wiki/new-design few things have not being updated intentionally as I dont want to do a giant patch.


Gozala:
I have being using this branch for a little while now and have not noticed any bugs so I feel more confident calling it done now.

PR #327, opened 2015-04-20
Make sure address-bar can't be in semi visible position.

### LINKED ISSUE #326: UI should never be rendered with half of the location bar visible
Today if you do following steps you end with a half visible location bar:

1. Load a page.
2. Scroll down so addressbar is no longer visible.
3. Type `Meta L` 

UI will end up in a state it should not be.

Gozala:
Fixes #326

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -51,7 +51,7 @@ body {
 main {
   height: 100vh;
   width: 100vw;
-  overflow-y: hidden;
+  overflow: -moz-hidden-unscrollable;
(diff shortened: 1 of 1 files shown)


Gozala:
@paulrouget  @gordonbrander can one of you do the honors ?

paulrouget:
I'm ok with this as a temporary workaround. But this is a gecko platform bug.

paulrouget:
(reverted. See #333)

PR #328, opened 2015-04-20
Move to webview implementation with records

Gozala:
This change is based of #324 and makes web view records based.


Gozala:
This also fixes #329 although I should probably fix that separately

PR #331, opened 2015-04-20
Semi working hotzone implementation.

Gozala:


gordonbrander:
@Gozala a couple of things:

- Hovering over the location bar causes the content to jump
- We should figure out a workaround for position-fixed elements inside of the page, but this could come later.

gordonbrander:
There is also a bug (platform?) which prevents you from moving the window via the window bar after it has faded in.

Gozala:
> There is also a bug (platform?) which prevents you from moving the window via the window bar after it has faded in.

Yeah it looks like there is a bug with `-moz-window-dragging` on non `relative` positioned elements.

paulrouget:
> Yeah it looks like there is a bug with -moz-window-dragging on non relative positioned elements.

I don't know exactly what's going here (haven't looked), but know that window-dragging regions are not updated as often as the layout for performance reason. We try to keep it as simple as possible. If we need window-dragging to work in this specific case and there's no way around it, I can update the layout code to take this into consideration.

paulrouget:
* Once the windowbar pops back because of the hotzone, we can't hide it by scrolling up. You have to scroll down first
* the content jumps when hovering the bar

paulrouget:
If we want to go for a not-perfect-but-good-enough solution, I would recommend a different approach: we could send an **async** message to the child process to request a **sync** message from the child process during a RFA, we smooth scroll the child process after the sync message, and smooth scroll the main process when we get the sync message.

I think this will ensure that both smooth scrolls are part of the same transactions.

There's still the IPC call from the child process that request the smoothScroll that might get some lag.

I don't know enough about how scroll operations are triggered from the child process (with or without async scroll), especially with smooth scroll operations.

We should try at least.

paulrouget:
With APZC, we can synchronize parent and child scroll operations. See this comment: https://github.com/mozilla/browser.html/pull/188#issuecomment-94635357

We're not far from being able to get that to work at the platform level, so let's make sure we don't spend too much time on this workaround.

If it stays straight forward, I'd be happy to land this PR (once reported comments are addressed).

Gozala:
I was unable to figure out why content was jumping I suspect it may be a platform bug. Also there is an issue with dragging so I'd rather not pursue it further.

PR #336, opened 2015-04-23
Kill flexbox model

paulrouget:
to review


paulrouget:
One of the issue I'm running into is the `order:n` property that is only support with the flex box model. Instead of trying to fix this, I'd prefer to remove the tab re-ordering. Any objection?

paulrouget:
Not for self: next, fix *fixmes*, then fix *order*, then merge.

paulrouget:
@Gozala can you please take a look at this?

I removed flexbox for everything except the dashboard and the tabstrip element (because of `order:n`).
I also moved some CSS code to JS.

gordonbrander:
Did a pretty thorough QA and can't find any bugs with the implementation. Going through the code now.

gordonbrander:
Note that this doesn't cover `dashboard.css`. OK with me. Better to do that in a separate commit.

PR #338, opened 2015-04-25
Various Jscs fixes

paulrouget:
Depends on #336

PR #340, opened 2015-04-27
Move banner code in its own components

paulrouget:
Depends on #338


gordonbrander:
Looks good. I have no easy way of QAing update mechanism, but code looks good and the browser runs without errors on my local branch.

Issue #346, opened 2015-04-30
Session restore regression


Gozala:
No longer seems to work

Issue #279, opened 2015-03-16
On some pages I get flickering scrollbars


Gozala:
This time I noticed it on github issues page but It's not consistent I think relates to the size of the page being almost the size available.

paulrouget:
Pretty sure it's #193.

PR #347, opened 2015-04-30
Fix session store regression #346

Gozala:
Fix for #346

### DIFF
--- a/src/browser/actions.js
+++ b/src/browser/actions.js
@@ -95,14 +95,17 @@ define((require, exports, module) => {
   // Reads stored session. Returns either immutable data for the
   // session or null.
   const readSession = () => {
+    const session = localStorage[`session@${version}`];
     try {
-      return fromJS(JSON.parse(localStorage[`session@${version}`]))
+      return session && fromJS(JSON.parse(session))
              .update('suggestions', Suggestions)
              .update('input', Editable)
              .update('webViews', WebViews)
-             .update('updates', Updates())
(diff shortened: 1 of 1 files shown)


src/browser/web-view.js
```diff
@@ -255,9 +255,9 @@ define((require, exports, module) => {
       onTitleChange: event => edit(WebView.setTitle(event.detail)),
       onPrompt: event => console.log(event),
       onAuthentificate: event => console.log(event),
-      onScrollAreaChange: event =>
-        edit(WebView.setContentOverflows(event.detail.height >
-                                         event.target.parentNode.clientHeight)),
+      onScrollAreaChange: !state.contentOverflows && (event =>
```
gordonbrander: Is this change related? Looks like a fix for scroll-away chrome.

Gozala: > Is this change related? Looks like a fix for scroll-away chrome.

No it was not intended here, good catch! It's for the other bug #279

src/browser/actions.js
```diff
@@ -95,14 +95,17 @@ define((require, exports, module) => {
   // Reads stored session. Returns either immutable data for the
   // session or null.
   const readSession = () => {
+    const session = localStorage[`session@${version}`]
```
gordonbrander: Nit: semicolon

paulrouget:
Doesn't that restore the update banner?

paulrouget:
See #352

Issue #193, opened 2015-03-05
Graphene scrolling freaks out on facebook.com


gordonbrander:
Open http://facebook.com, make sure you're logged out. This generally seems to happen to me when the window is expanded to take up the full screen of my Mac Book Pro 16".

cc @paulrouget

paulrouget:
It's a frontend bug, not a Graphene bug. It happens because we update `contentOverflows` more than once, and some pages change their layout on resize. So we should allow toggling `contentOverflows` only once.

Gozala:
@paulrouget I see what you saying, but I'm not sure proposed solution is actually going to work:

- Single page apps like gmail fetch data on clicks and replace displayed UI, in turn you may initially set `contentOverflows` to `false` but then you may need to set to `true` if we drop those updates it's not going to work correctly for such websites.
-  I had other case in mind that I forgot while writing first one :(

Gozala:
Ok so #348 does what @paulrouget suggested in the comment, I do not think it's ideal (see my previous comment) still better than what we have now though, so I think we can go with it for now.

paulrouget:
This is not supposed to happen anymore.

PR #348, opened 2015-05-01
Fix flickering scrollbars #193

Gozala:
Workaround for #193

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -71,7 +71,7 @@ define((require, exports, module) => {
     canGoBack: Boolean(false),
     canGoForward: Boolean(false),
 
-    contentOverflows: Boolean(false),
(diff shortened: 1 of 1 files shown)


gordonbrander:
This fix does not work for me (checked out branch, reset session, refreshed browser, restarted browser, loaded facebook.com). I still get the flickering scrollbar.

gordonbrander:
NVM, this was human error when checking out the branch. Sorry for the false alarm.

paulrouget:
Why do you use an undefined value? Set it to `false` by default and only allow it to go from `false` to `true`. With this setting, you only listen to the first event. We should only allow one toggling.

We want this to work: `no overflow -> no overflow -> overflow`.

paulrouget:
See #353

PR #352, opened 2015-05-04
do not store update state

paulrouget:


paulrouget:
@Gozala r?

src/browser/actions.js
```diff
@@ -114,6 +114,10 @@ define((require, exports, module) => {
       .setIn(['rfa', 'id'], -1)
       // Reset state of each web viewer that can't be carried across the sessions.
       .updateIn(['webViews'], viewers => viewers.map(WebView.persistent))
+      .updateIn(['updates'], updates => updates.merge({
```
Gozala: You could just do `.remove('updates')` instead. But this works too.

PR #353, opened 2015-05-04
toggle contentOverflows only once

paulrouget:


paulrouget:
@Gozala r?

paulrouget:
Fix #193

src/browser/web-view.js
```diff
@@ -260,7 +260,11 @@ define((require, exports, module) => {
       onTitleChange: event => edit(WebView.setTitle(event.detail)),
       onPrompt: event => console.log(event),
       onAuthentificate: event => console.log(event),
-      onScrollAreaChange: state.contentOverflows == void(0) && (event =>
+      // This will trigger a resize. If the content react to the resize by changing its
+      // layout, this might change the scrollarea again, triggering a resize… infinite
+      // loop.
+      // So we only allow contentOverflows to transition from false (default value) to true.
+      onScrollAreaChange: !state.contentOverflows && (event =>
```
Gozala: I tried this approach first and observed odd behavior on gmail.com and few other sites, I was getting double scrolling or something strange like that. I'm unable to test this right now, but please test on gmail.com and see if that really a case we should probably not land this change.

paulrouget: I tried and can't reproduce.

If there's anything strange happening, the only thing I could see is a weird behavior when we reset `contentOverflows` on load.

Gozala:
It is really strange, I just reproduced this last night:

1. Load gmail
2. Login
3. Scroll to the bottom
4. Scroll up.

You'd see nabber scroll out on 3. and on 4. you'd see both gmail email list scrolling and nav bar scrolling into the view as well which felt not smooth but really bad.

That being said I'm no longer able to reproduce this and navbar no longer scrolls out on gmail for me, it could be that they updated something over night ? Anyhow as no one can reproduce it we should probably just land it and keep an eye if we notice something along the lines I've described.

Gozala:
@paulrouget you should bump a version though as restoring a session fails given that stored `contentOverflows` used to be `null` which is invalid value for `Boolean` field.

Gozala:
@paulrouget I was wrong when I said I'm unable to reproduce, I actually was not running browser.html with your changes which is why I was unable to reproduce. I posted recording of the issue in the slack. No idea why you or @gordonbrander can't seem to reproduce it.

gordonbrander:
We had a super long discussion about this on Slack and it looks like this is a scroll syncing issue that is worse on @Gozala's machine because of platform performance. I'm wondering if https://github.com/mozilla/browser.html/issues/355#issuecomment-98943468 would help here too?

gordonbrander:
> gozala [12:30 PM]
> @gordonb: no it’s just sites that resize later
> @gordonb: problem is we get an event if content overflows or not
> and then in my case we stick to that, but page later on could resize needing to overflow
> in which case we would allow navbar scrolling

paulrouget:
I don't understand what the issue is. Isn't that behavior always been here? Isn't that exactly what scrollgrab is supposed to do? There is some issues with scrollgrabs because we don't use APZC (parent and child scroll at the same time), but that will get fixed soon. Is that what's bothering you? This behavior existed before.

paulrouget:
@Gozala ping?

Gozala:
@paulrouget we had a lengthy discussion about this with @gordonbrander on slack & we even posted some videos of gmail and other sites. As it tuns out it is expected behavior it just felt really odd on my eyes on some sites especially gmail. On other sites it feels like you scroll navbar first and then the page while on gmail they both seem to scroll in parallel which made it feel odd to me.

In conclusion I believe we wanted to land this change, sorry for not posting back that here.

I'm glad APZC is going to resolve this ;)

PR #354, opened 2015-05-04
command line handler

### LINKED ISSUE #342: browser.html as default browser can't handle links
So today I've set browser.html as my default browser. Now if I click links in my email client I see it reloading active tab and that is it. Have not digged into what's going on but we should figure it out.

paulrouget:
This is not working for now. Depends on Part 8 of bug 1134106. Will fix #342.


Gozala:
Looks good to me.

PR #357, opened 2015-05-05
basic netError support

### LINKED ISSUE #286: support about:netError
We should show an error page. I would prefer to handle the error page in HTML as we do with everything else.

paulrouget:
WIP. Depend on https://bugzilla.mozilla.org/show_bug.cgi?id=1161435.

Fix #286

### DIFF
--- a/src/netError/netError.html
+++ b/src/netError/netError.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+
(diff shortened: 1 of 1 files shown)


paulrouget:
Landed a basic error page: 52a5b14ee24ca0feb80fa63aebcb5f3d43cc3465

PR #373, opened 2015-05-14
History db

Gozala:
This not complete and scoring of results can definitely be improved, but it's a start. This also not yet stores tile images into DB.


src/browser/browser.js
```diff
@@ -21,7 +21,7 @@ define((require, exports, module) => {
   const {WebViewBox, WebView} = require('./web-view');
   const {Dashboard} = require('./dashboard');
   const {readDashboardNavigationTheme} = require('./dashboard/actions');
-  const {activate: activateStrip, readInputURL,
+  const {activate: activateStrip, readInputURL, sendEventToChrome,
```
paulrouget: ?

src/browser/history.js
```diff
@@ -0,0 +1,320 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
paulrouget: Can you move this file to /common/ ?

src/lang/task.js
```diff
@@ -0,0 +1,49 @@
...
+          then(resume);
+        }
+      };
+
+      resume(next());
+    });
+  };
+  exports.spawn = spawn;
+
+  const async = routine => function(...params) {
```
gordonbrander: ...so this is an ergonomic promise wrapper around generators?

src/browser/browser.js
```diff
@@ -168,6 +169,32 @@ define((require, exports, module) => {
     position: 'relative'
   };
 
+
+  // History API hooks
+  const history = new History({withTop: true});
```
gordonbrander: Something more descriptive? Maybe `withTopSites`?

src/browser/browser.js
```diff
@@ -168,6 +169,32 @@ define((require, exports, module) => {
...
+
+  const changeTitle = ({webView, title}) => {
+    history.edit(Site.from(webView), site => site.set('title', title));
+  };
+
+  const changeImage = ({webView, image}) => {
+    history.edit(Site.from(webView), site => site.set('image', image));
+  };
+
+  const changeIcon = ({webView, icon}) => {
```
gordonbrander: Should these 3 things be a single function? That way we could batch changes if we were changing more than one thing at once.

src/browser/history.js
```diff
@@ -0,0 +1,320 @@
...
+           index + 1 == xs.size ? xs.butLast() :
+           xs.take(index).concat(xs.skip(index + 1))
+  }
+  const include = x => xs => xs.indexOf(x) < 0 ? xs.push(x) : xs
+
+
+  /*
+  ## Interests API
+
+  ### Sites
```
gordonbrander: This should probably be called `pages` or `locations`, not `sites`. Sites are root domain, like `facebook.com`. I think we will want to surface only root domains in some places like favorites tiles, so it would be better if we got the terminology right up front.

src/browser/history.js
```diff
@@ -0,0 +1,320 @@
...
+      "functional"
+    ]
+  }
+  */
+
+  const Visit = Record({
+    type: Type('Visit'),
+    start: TimeStamp,
+    end: Maybe(TimeStamp),
+    id: String, // This is a web view ID in our case.
```
gordonbrander: should we call this `webViewID`?

src/browser/history.js
```diff
@@ -0,0 +1,320 @@
...
+
+  Site.endVisit = ({id, time}) => site => {
+    const index = site.visits.findIndex(visit => visit.id === id);
+    return index < 0 ? site : site.setIn(['visits', index, 'end'], time);
+  };
+
+
+/**
+
+### Quotes
```
gordonbrander: Nit: I would like to rename to highlights. NBD.

src/browser/history.js
```diff
@@ -0,0 +1,320 @@
...
+
+Quotes store contains records of quotes that have being created by a user that
+have a following structure.
+
+
+{
+  _id: "quote/W29iamVjdCBBcnJheUJ1ZmZlcl0=",
+  content: `If you say that <span class="fixed">a</span> is 5, you can't say it's something else later because you just said it was 5. What are you, some kind of liar? So in purely functional languages, a function has no side-effects. The only thing a function can do is calculate something and return it as a result.`
```
gordonbrander: I think we also want to capture some of the surrounding text before/after the highlight so we can show context.

src/browser/browser.js
Gozala: Artifact of my initial attempt on #32b17d7bd92ad1825f19fa0431815380ffb7f813

Good catch

src/browser/browser.js
Gozala: We get separate DOM events for each of them and one of them may not even happen, so not worth it.

src/browser/history.js
Gozala: makes sense.

src/browser/history.js
Gozala: Can we leave quotes / highlights as is until we actually start using them, I'd like to avoid regressions by renaming stuff. I did manual tests on them and renaming would require doing manual testing again as we otherwise won't spot errors as it's not used yet.

src/browser/history.js
Gozala: Can we leave it as is for now as it would really depend on implementatiot details, like image can be captures as base64 uris for example or we may need to capture styles as well, etc...

I'm happy to change this and rename once we start worknig on it. I just included it so I don't loose implementatino that I did back when it was `firefox.html`.

gordonbrander: :+1:

src/browser/history.js
```diff
@@ -0,0 +1,321 @@
...
+        include_docs: true
+      });
+    }
+    setupListeners() {
+      this.topPageChangeFeed.on("change", this.onTopPagesChange);
+      this.pagesChangeFeed.on("change", this.onPageChange);
+    }
+
+    // Edits per record are queued, to avoid data loss
+    // due to concurrent edits.
```
gordonbrander: When, if ever, do we have concurrent edits on the client side?

src/browser/history.js
```diff
@@ -0,0 +1,321 @@
...
+      const store = history.stores[record.constructor.storeID];
+      return edit(record, transform, store);
+    }
+  })
+
+  // History
+  class History {
+    static defaults() {
+      return {
+        pagesStoreName: "pages",
```
gordonbrander: Nit: double quotes all over the place. We tend to prefer `'` to `"`.

src/browser/history.js
```diff
@@ -0,0 +1,321 @@
...
+      const {pagesStore, quotesStore, tagsStore,
+             pagesStoreName, quotesStoreName, tagsStoreName} = this.options;
+
+      this.stores = {
+        pages: pagesStore || new PouchDB(pagesStoreName),
+        quotes: quotesStore || new PouchDB(quotesStoreName),
+        tags: tagsStore || new PouchDB(tagsStoreName)
+      }
+
+      this.editQueue = Object.create(null);
```
gordonbrander: Nit: there might be a reason for this that I am missing, but is `null` prototype actually better here than `{}`?

src/browser/history.js
Gozala: > When, if ever, do we have concurrent edits on the client side?
> 
> Also, doesn't supplying a _rev cause Pouch to automatically handle these things for us? Seems like this is exactly what Couch/Pouch are designed to do. Please feel free to tell me I am wrong if I'm missing something here.

Here is an example:

1. Load `facebook.com`
2. Create `facebook.com` page record.
3. Get title change event edit that records `title`.
4. Get change icon event edit records `icon`.

Last three steps make edits to the same record and 4th edit may start before 3rd is complete in which case both 3rd and 4th operation will refer to the `_rev` from 2nd edit and there for run into conflict and get rejected.

In practice though we load many tabs at the same time and some of them maybe for the same page and even if not each edit triggers update on `top/sites`. So different page record do get updated concurrently with `top/sites` and there for without queueing edits you

src/browser/history.js
```diff
@@ -0,0 +1,321 @@
...
+      this.onTopPagesChange = this.onTopPagesChange.bind(this);
+
+      this.options = Object.assign(History.defaults(), options);
+      const {pagesStore, quotesStore, tagsStore,
+             pagesStoreName, quotesStoreName, tagsStoreName} = this.options;
+
+      this.stores = {
+        pages: pagesStore || new PouchDB(pagesStoreName),
+        quotes: quotesStore || new PouchDB(quotesStoreName),
+        tags: tagsStore || new PouchDB(tagsStoreName)
```
gordonbrander: This may be a work around for lack of PouchDB views? Just thought I would call out that Pouch is designed to be used in such a way that you dump everything in one database and create view indexes for your different "buckets".

src/browser/history.js
Gozala: Oh yeah I wrote most of this code before we decided to use single quotes over doubles.

src/browser/suggestion-box.js
```diff
@@ -190,18 +193,94 @@ define((require, exports, module) => {
...
+      const count = text.length;
+      const match = pattern.exec(text);
+
+      return !match ? -1 :
+              base +
+              length * Math.sqrt(match[0].length / count) +
+              index * (1 - match.index / count);
+  }
+
+  const Pattern = (input, flags="i") => {
```
gordonbrander: Maybe pattern should be part of `lang` or `common`?

src/browser/suggestion-box.js
```diff
@@ -190,18 +193,94 @@ define((require, exports, module) => {
...
+                  // and match index.
+                  const titleScore = score(query, page.title);
+                  const uriScore = score(query, page.uri);
+
+                  // Store each score just for debuging purposes.
+                  page.frequencyScore = frequencyScore;
+                  page.titleScore = titleScore;
+                  page.uriScore = uriScore;
+
+                  // Total score is ranked form `-1` to `1`. Score is devided into
```
gordonbrander: Idea: would math be easier if score was `0..1` rather than `-1..1`?

gordonbrander:
@Gozala I get a white screen on this branch, which I'm guessing has to do with the change from `localStorage` to PouchDB.

We need to make the switch without requiring you to delete profile. We can trash all the old data though.

src/browser/history.js
Gozala:
```js{}[__proto__]

src/browser/history.js
Gozala: Yes & No. Yes becaues we can't don't have eval so we can't make views. No because views are not magical and they do have an overhead, like db needs to actually keep indexing meaning that once you inject a record into a db it needs to through your view functions then create link records in so they can refer to actual records in this db. During seach same story it needs to lookup by index first then get by id.

What I'm trying to say is there is no real benefit of storing everything in one DB, it's actually even worse given the way we are doing a search. While using separate bases don't really have any downside to them.

src/browser/suggestion-box.js
Gozala: > Idea: would math be easier if score was 0..1 rather than -1..1?

Negative values are there so we that score of individual field can actually have a negative impact of overal score. Probably it still could be made to be in 0 to 1 range but seems less obvious to me.

Gozala:
> @Gozala I get a white screen on this branch, which I'm guessing has to do with the change from localStorage to PouchDB.
> 
> We need to make the switch without requiring you to delete profile. We can trash all the old data though.

No you just need to run `npm install` to get pouch.

src/browser/history.js
gordonbrander: :+1:

src/browser/history.js
gordonbrander: Ok, is this something we could solve by adding a conflict resolution function? http://pouchdb.com/guides/conflicts.html

src/browser/history.js
gordonbrander: I get that, but why does it matter?

src/browser/history.js
gordonbrander: I'm wondering if it will have downstream consequences for how we sync to Couch. With one database, sync story is very simple.

src/browser/suggestion-box.js
gordonbrander: wfm

gordonbrander:
Oops, sorry.

src/browser/history.js
Gozala: That's a best practice if you're using object as a hash map in JS. I'll add the comment.

src/browser/history.js
Gozala: I'll look into maybe using single DB instead with a help of `startkey` we may get away without perf penalty.

src/browser/history.js
Gozala: So you do need to do both, issue with an "upsert" appreach is that it will keep re-trying same edit over & over unless it hits limit of retryies or succeeds. Which is waste of recosurce when we can just queue up because we do know we'll have conflict if we don't wait for completion.

Now I do agree you always need to handle conflicts but in distributed system, which is not what we have yet. In other words I was planning to deal with it once we'll start syncing with backend. Until than it's not really going to do anything.

Gozala: Anyway I don't mind adding retry a.k.a "upsert" on conflicts althugh we still should  keep the queing to avoid wasting resources.

gordonbrander: Ok, wfm. Just wanted to make sure we weren't building machinery that Pouch already has.

src/browser/suggestion-box.js
Gozala: > Maybe pattern should be part of lang or common?

It would make sense to add `lang/parse` or somethig but can we leave it here for now, until we need it elsewhere. Just trying to avoid explosion of modules that we need to fetch before start an app.

gordonbrander: wfm

gordonbrander:
Yup, having history completion is awesome. Pretty much all I need.

Issue #380, opened 2015-05-18
Fresh profile seems to be created every now and then


Gozala:
Now that we have browser history integrated loosing a profile became a very irritating issue, as I keep loosing my history info every now and then.

Gozala:
As far as I can see running browser.html with `npm start` and then running by starting `b2g` or other way round reliably looses my profile data. Could it be related to github serving content through `https` and local server through `http` ?

fabricedesre:
for sure http and https are different origins, so indexedDB data ends up in different cookie jars.

paulrouget:
It's not only http vs. https, the domain is also different.

@Gozala Is it really a new profile or just the db content that is different?

paulrouget:
Apparently, it's because the db is been deleted everytime the app is being re-installed (switching between origin triggers a re-install).

@Gozala mentioned using a different app name for dev profile.

I think we should just use a different profile.

@Gozala what do you think?

Gozala:
I have few constrains with this:

1. It won't be visible to make something as simple as `npm start` which is unfortunate as it would make contribution more difficult. Although not a big deal right now.
2. Larger constraint is that graphene seem be too bound to a single app, meaning if I run any other app say app that "vr" team is working on I'll loose data of both apps.

I think fixing this is effort well spend especially if we hope graphene to be useful outside of this project. Using separate profile in the meantime is fine.

Not sure where this fall in terms of priority scale though, I'll let @gordonbrander speak of that :)

paulrouget:
We can fix this. But really not a priority compared to all the other things.

Gozala:
BTW another unfortunate thing I noticed which is kind of relates to this bucket is that it's impossible to run multiple graphene based apps at a time (without specifying profile).

paulrouget:
Not sure to understand why. Graphene can run multiple apps. It's based on B2G, and B2G can run multiple apps.

Gozala:
> Not sure to understand why. Graphene can run multiple apps. It's based on B2G, and B2G can run multiple apps.

@paulrouget Just create any other app but **browser.html** say **webvr** start it, first you'll get a see a message that `b2g` is already running. Another step would be to quit **browser.html** and start **webvr** with will work  but then you loose all the data stored by **browser.html**.

[...]

Issue #306, opened 2015-04-06
`npm start` starts server, but doesn't stop it


paulrouget:
STR:
* `npm start`
* quit browser
* `ps aux|grep npm` shows a running instance of *ecstatic*
* `npm start` and `npm run start-server` will throw

Gozala:
It does stops it if you `^ c` in the terminal but does not if you quit b2g. Not sure exactly that is a case to be honest. Currently `npm start` does `npm run start-server & npm run start-app` is there a way to run two processes such that exiting one would exit other as well ?

Only other thing I can think of is to do `npm run start-server | npm run start-app` instead but then output from first process is lost which could be ok though.

Gozala:
Fixed by #381

PR #381, opened 2015-05-20
Use custom profile when running in dev mode

Gozala:
Mitigate #380 by using local profile when running with `npm start`. Also address #306


paulrouget:
Looks good. Can you keep `start-server` for people (like me) who run graphene without npm?

PR #389, opened 2015-05-28
add new tiles and update tiles

mmhyamin:
1. New tiles of ebay.com, microsoft.com, paypal.com, vk.com added.
2. Updated facebook.com, youtube.com and wordpress.org's official logo tiles.


gordonbrander:
@mmhyamin these are great, excepting I see no reason to change the WordPress tile.

mmhyamin:
@gordonbrander [wordpress.com](https://wordpress.com) is a blog hosting service by [Automattic](http://en.wikipedia.org/wiki/Automattic). They use the blue background logo. But [wordpress.org](https://wordpress.org/) is the site of [WordPress CMS by WordPress Foundation](http://en.wikipedia.org/wiki/WordPress). Most of the time they use the ash color logo. If we use the same logo for both site it will confuse others which one is wordpress.org or which one is wordpress.com.

Gozala:
@gordonbrander what are we doing with this pull request ?

gordonbrander:
This is useful work. The new design we're working on moves away from tile images, but we may bring them back once we can figure out a good way to get high-res images from non-hardcoded sites.

I think the best thing will be to rebase this pull request at a later date when we take another look at tiles.

PR #392, opened 2015-06-03
Basic integration test support.

KevinGrandon:
This is using some forked b2g testing packages here: https://github.com/kevingrandon/graphene-marionette-runner

It's still a bit rough around the edges, but should give you an idea of how things could work. Next up would be importing the node http servers from gaia to serve static content from a fixtures directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+graphene
 .DS_Store
 node_modules/
 .profile/
--- a/README.md
+++ b/README.md
@@ -23,3 +23,11 @@ In the future, we want `browser.html` to be able to run on top of Servo.
 
 The easiest way to use developer tools with Browser.html is to select the "Remote Runtime" option in WebIDE.
 By default you should be able to connect to the running browser at: localhost:6000.
+
+
+### Integration Tests
+Run integration tests with `./test/runall.sh`. You need to have a graphene gecko binary symlinked in the root of the repository.
+
+```
+ln -s ../gecko/obj-graphene/dist/Graphene.app graphene
+```
--- a/package.json
+++ b/package.json
@@ -29,6 +29,9 @@
   ],
   "dependencies": {
     "immutable": "git://github.com/jcranendonk/immutable-js.git#0986651df9398187390f4638424adb532985439f",
+    "marionette-client": "^1.7.5",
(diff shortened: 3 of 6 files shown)


gordonbrander:
@Gozala can you review?

paulrouget:
(awesome!)

paulrouget:
(I'll look at this)

paulrouget:
I can't get this to work:

```sh
KevinGrandon-testing > ./test/runall.sh
#!/bin/bash -ve

./node_modules/.bin/marionette-mocha \
  --harmony \
  --host-log stdout \
  --runtime ./graphene/Contents/MacOS/graphene \
  $(find test -name '*_test.js') $@;
find test -name '*_test.js'


  ․․

  0 passing (2s)
  2 failing

  1) new tab test "before each" hook:


      + expected - actual


  rror: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

  2) new tab test "after each" hook:

      Uncaught
      + expected - actual


  ypeError: Cannot read property 'send' of undefined
```

KevinGrandon:
Hmm, I did get the tests to pass once, but not anymore, heh. I think I need to fiddle some prefs, so might need a gecko patch here as well.

KevinGrandon:
@paulrouget - Tweaked it and the test appears to work for me. Can you try the latest version of this branch and let me know what you think?

paulrouget:
It does nothing.

I see this and it hands there:

```
#!/bin/bash -ve

./node_modules/.bin/marionette-mocha \
  --host-log stdout \
  --host $(pwd)/node_modules/graphene-marionette-runner/host/index.js \
  --runtime ./graphene/Contents/MacOS/graphene \
  --start-manifest http://localhost:6060/manifest.webapp \
  $(find test -name '*_test.js') $@;
pwd
find test -name '*_test.js'
```

Also, I had to prefix `marionette-mocha` with `./node_modules/.bin/`

KevinGrandon:
Weird! I'll look into it - which OS are you using? Looking at the patch, it probably only works on mac right now. Will try to get some cross platform fixes in.

paulrouget:
osx.

paulrouget:
r+. We figured out what was going on (fix about to land in larch).

PR #405, opened 2015-06-18
devtools hud

paulrouget:


src/browser/devtools.js
```diff
@@ -0,0 +1,121 @@
...
+  const {DOM} = require('react');
+  const Component = require('omniscient');
+  const {Record, List} = require('typed-immutable/index');
+
+  const DevTools = Record({
+    enableHUD: Boolean(false),
+    enableRemoteDevtools: Boolean(false),
+    console: Record({
+      // FIXME: I don't how to update these list from anywhere in our codebase.
+      log: List(String),
```
Gozala: Lists implement Array interface and you can use same `update` and `set` stuff as with other structures. For example you could do:

```js
const log = (devtools, message) =>
  devtools.set('console', devtools.console.push(message)
```

paulrouget: The problem is not that. The problem is that at anytime in the code, and maybe not in a render function, I want to be able to update the state.

Basically, in devtools.js, I'd like to be able to do something like:

```js
console.log = (...args) => {
  realConsole.log(...args);
  devtools.set('console', devtools.console.push(message));
}
```

but I don't have access to an edit function.

Gozala:
> It's annoying that I have to add the VirtualEvent inside main.js. Is there a simple way to add the mozSettings listener inside devtools.js?

We kind of can do it already by calling `step` on redererer, but it's odd. In the my rewrite exposing additional sources of input will be simple and natural.

> I would like to intercept all the calls to console.log/error to show them in the HUD too. But to do so, I need to be able to update the state from anywhere in our code base. How can I do that?

I don't think you need to be able to update from anywhere you would just need to do it from the definition sight of the `console.log / error`. Here you can do it again via `renderer.step(state => state)` or better wait for my changes to land as it would be a lot more elegant there.

Gozala:
I am little concerned by the fact that app will save every message ever logged, I think that would have quite visible impoct on performance and memory footprint.

Can add those settings to `about:settings` instead or maybe to `about:devtools`

src/browser/devtools.js
```diff
@@ -0,0 +1,121 @@
...
+      error: List(String)
```
Gozala: You can log objects so this does not seems right, you could do `List(Any)` btw.

paulrouget:
> better wait for my changes to land as it would be a lot more elegant there.

Ok!

> I am little concerned by the fact that app will save every message ever logged, I think that would have quite visible impoct on performance and memory footprint.

We can start intercepting logs only when the HUD shows up. Or we can just not have a console in the HUD. I just thoughts that would come handy.

paulrouget:
I've updated the PR. There's a FIXME in there I don't know how to fix.

PR #419, opened 2015-06-29
initial animations for previews

paulrouget:


src/browser/web-view-deck.js
```diff
@@ -202,10 +202,29 @@ define((require, exports, module) => {
...
+    let scaleY = 1;
+    let opacity = 1;
+    let pointerEvents = 'auto';
+
+    let display = selected && !selected.getIn('shell', 'isFocused');
+
+    if (!isActive) {
+      opacity = 0;
+      scaleX = 240 / innerWidth;
+      scaleY = 300 / (innerHeight - 56); // 56: size of the window bar
```
Gozala: Use of `window.innerWidth` and `window.innerHeight` like this is not good. In general `view` is not supposed to use any mutable references that were not explicitly passed to it or it would cause issues. In this specific case resizing a window will not cause this `view` and there for numbers used in here will no longer be valid.

paulrouget:
Not relevant anymore.

PR #450, opened 2015-07-08
add transitions and gestures to cards

paulrouget:
This is the first step. Transitions/animations are not properly tweaked, but it should work for now, and perf should be good.


paulrouget:
I removed preview-box.js because it is not used.

src/browser/web-view.js
```diff
@@ -286,7 +286,13 @@ define((require, exports, module) => {
...
       key: 'web-views',
       style: {
-        transform: `scale(${isActive ? 1 : 0})`
+        width: '100vw',
+        height: '100vh',
+        transition: `transform ${isActive ? 90 : 300}ms linear, opacity 150ms linear`,
+        transform: `scale(${isActive ? 1 : 0})`,
+        opacity: isActive ? 1 : 0,
+        pointerEvents: isActive ? 'initial' : 'none',
+        transformOrigin: `${selected * 260 + 130}px 50%`, // Center of the selected card
```
Gozala: Could you please instead add two entries to the style sheet maybe `zoomIn` & `zoomOut`  and assign either one or another based on `isActive` ? Alternatively you could also factor out common parts and to `Style(style.webViews, isActive ? style.zoomIn : style.zoomOut)` to reduce number of allocations.

PR #453, opened 2015-07-08
Use CSS to animate progressbars

paulrouget:
Instead of having one single progressbar, here, I build one progressbar per webview. That allows us to keep the `translateX` position when switching tabs. With some transition and animation magics, we can reproduce the previous behavior.

The result is a much smoother animation.

Maybe the progresssbar should be built outside of the webview.

I'm wondering if passing the theme into the webview then to the progressbar is a problem (iirc, we had some issues because of that earlier).

I renamed Progress to WebProgress just to make my life easier when doing the rewrite. We can keep it that way or go back to Progress(bar).

### DIFF
--- a/src/browser/progress-bar.js
+++ b/src/browser/progress-bar.js
@@ -6,164 +6,125 @@ define((require, exports, module) => {
 
   'use strict';
 
-  const {html} = require('reflex');
-  const {animate} = require('common/animation');
   const {Record, Union} = require('common/typed');
-  const Theme = require('./theme');
   const {StyleSheet, Style} = require('common/style');
+  const {html, render} = require('reflex');
 
   // Model
-
   const Model = Record({
-    loadStarted: 0,
-    connected: 0,
-    loadEnded: 0,
-
-    value: 0
+    loadStarted: false,
+    loadEnded: false,
   });
   exports.Model = Model;
 
-  // Returns `true` if associated web-view is loading a page.
-  const isLoading = ({loadStarted, value}) =>
-    loadStarted > 0 && value < 1;
-  exports.isLoading = isLoading;
-
-  // Returns `true` if associated web-view connected to server.
-  const isConnecting = ({loadStarted, connected}) =>
-    loadStarted > 0 && connected <= 0;
-  exports.isConnecting = isConnecting;
-
-
-  // Actions
-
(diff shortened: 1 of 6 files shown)


src/browser/web-progress.js
```diff
@@ -0,0 +1,122 @@
...
+  }, 'WebView.WebProgress.LoadEnd');
+
+  const Action = Union({LoadStart, LoadEnd});
+  exports.Action = Action;
+
+  const update = (state, action) =>
+    action instanceof LoadStart ? state.merge({
+      loadStarted: true,
+      loadEnded: false
+      }) :
```
Gozala: Nit: Can we please move **then** block to the next line when it does not fit on the same line to match code elsewhere, so this would become:

```js
action instanceof LoadStart ?
  state.merge({loadStarted: true, loadEnded: false}) :
action instanceof LoadEnd ?
  state.merge({loadEnded: true})
state;
```

Gozala:
> Maybe the progresssbar should be built outside of the webview.

I would suggest to move progress bars outside of the web view similar to how `web-previews` are rendered right now. That way views can be updated independently of each other.

src/browser/progress-bar.js
```diff
@@ -6,164 +6,125 @@ define((require, exports, module) => {
...
 
-    // If `webView` is loading then animate node with `ProgressChange` actions
-    // on every animatation frame.
-    return !progress ? null :
-           !isLoading(progress) ? node :
-           animate(node, address.pass(ProgressUpdate, id));
+  const view = (loaders, progress, selected, theme) => {
+    return html.div({
+      key: 'progressbars',
+      style: Style(containerStyle.base)
```
Gozala: Nit: No need to wrap single item in `Style` it's just for composing several of them.

src/browser/web-view.js
```diff
@@ -237,7 +237,6 @@ define((require, exports, module) => {
     return IFrame.view({
       id: `web-view-${loader.id}`,
       uri: location,
-      className: `web-view ${isSelected ? 'selected' : ''}`,
```
Gozala: Nit: I found it useful with an devtools inspector to identify which one is selected.

Gozala:
r+

paulrouget:
comments addressed.

PR #454, opened 2015-07-08
Further improvements to preview cards

gordonbrander:
More work on cards and scraper

### DIFF
--- a/src/browser/web-preview.js
+++ b/src/browser/web-preview.js
@@ -97,13 +97,15 @@ define((require, exports, module) => {
       whiteSpace: 'nowrap'
     },
     icon: {
+      backgroundSize: 'cover',
+      backgroundPosition: 'center center',
+      backgroundRepeat: 'no-repeat',
       borderRadius: '3px',
       position: 'absolute',
       right: '4px',
       top: '4px',
       width: '16px',
       height: '16px',
-      MozForceBrokenImageIcon: 0
     },
     image: {
       backgroundColor: '#DDD',
--- a/src/browser/web-preview.js
+++ b/src/browser/web-preview.js
@@ -66,7 +66,7 @@ define((require, exports, module) => {
       borderRadius: '4px',
       boxShadow: '0 1px 3px rgba(0, 0, 0, 0.4)',
       color: '#444',
-      display: 'inline-block',
+      float: 'left',
       height: '300px',
       margin: '0 10px',
       overflow: 'hidden',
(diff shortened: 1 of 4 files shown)


src/browser/web-preview.js
```diff
@@ -223,26 +225,40 @@ define((require, exports, module) => {
...
       position: 'absolute',
       top: 0,
       zIndex: 0,
-      MozWindowDragging: "drag"
+      MozWindowDragging: 'drag',
+    },
+    previews: {
+      // This is important. We need previews to make space around itself.
+      // Margin doesn't play well with scroll -- the right-hand edge will get
+      // cut off, so we turn on the traditional CSS box model and use padding.
```
paulrouget: That's surprising. The right margin doesn't account for the scroll area? It should.

src/browser/web-preview.js
```diff
@@ -223,26 +225,40 @@ define((require, exports, module) => {
...
+      boxSizing: 'content-box',
+      width: '100vw',
+      // Fixed height to contain floats.
+      height: '300px',
+      padding: 'calc(100vh / 2 - 150px) 100px 0 100px',
```
paulrouget: `100vh / 2` -> `50vh`

src/browser/web-preview.js
```diff
@@ -223,26 +225,40 @@ define((require, exports, module) => {
...
-    preview: {
-      width: '100vw',
-      height: '100vh',
-      paddingTop: 'calc(100vh / 2 - 150px)',
+    scroller: {
       backgroundColor: '#273340',
-      overflowX: 'auto',
+      height: '100vh',
+      width: '100vw',
+      overflowY: 'hidden',
```
paulrouget: How can it overflow vertically?

paulrouget:
What about scroll snapping?

src/browser/web-preview.js
gordonbrander: Margin has always had problems interacting with scrollable wrappers, in all browsers.

We use `float:left` because `display:inline-block` renders additional gaps in between elements when space or carriage returns are present in the source code. Anyway, cards are a fixed height, so we can just fix the container height.

gordonbrander:
No scroll snapping yet. We'll look at that later.

paulrouget:
Instead of setting the width manually, can you use non-wrapped inline-block elements and set the font-size to 0 to avoid the spacing?

src/browser/web-preview.js
gordonbrander: Whoops! Inherited this from old code. Will fix.

gordonbrander:
@paulrouget I don't like to take this approach because it affects the fonts of the elements inside.

paulrouget:
How so? Can't you set the the font-size to 0 and then reset it to the proper font-size? Also, using inline-block will allow you use vertical alignment instead of setting a dependency to the card height, and you won't have to update the scroller width every time.

Anyway, I don't want to argue about CSS implementation at this stage. Feel free to land your PR.

Everything will be more straight forward the day we can use the flexbox model.

gordonbrander:
Resetting is what I mean. `font-size` is inherited, where `float` is not, so fewer surprises. Also, in future, we'll want to have essentially limitless tabs. In that case we'll want to set the width of the container element if we want to do lazy rendering.

gordonbrander:
Yeah I hear what you're saying. In this specific case there are design reasons I wouldn't want to have different card heights, so it's a specific case. I guess it's NBD either way.

magsout:
>How so? Can't you set the the font-size to 0 and then reset it to the proper font-size? Also, using inline-block will allow you use vertical alignment instead of setting a dependency to the card height, and you won't have to update the scroller width every time.

Or maye you can use, `display:flex; align-items: center;` (to the parent) instead of `float` or `display:inine-block` 

Fast and scalable

paulrouget:
We are trying to stay away from the flexbox model for now. Not supported by Servo yet.

magsout:
Ok, make sense ;)

PR #463, opened 2015-07-10
Synthesis interaction

Gozala:
- Wire up zoom in / out gestures.
- Wire up new tab button.
- No longer switch from create to edit mode to match spec.

### DIFF
--- a/src/browser/synthesis-ui.js
+++ b/src/browser/synthesis-ui.js
@@ -11,6 +11,7 @@ define((require, exports, module) => {
   const WebView = require('./web-view');
   const Input = require('./web-input');
   const Suggestions = require('./suggestion-box');
+  const Gesture = require('service/gesture');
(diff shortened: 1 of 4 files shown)


gordonbrander:
Going to log interaction bugs here. I know some of this stuff is simply not implemented, but want to capture it.

- In new tab mode `cmd w` will close the new tab and the previously open tab.
- Clicking cards does nothing
- Clicking search suggestions does not launch page
- Numerous bugs with the color matching code. For example, focusing/unfocusing the url bar will cause the window controls to change color.
- Cards should be ordered LTR, newest -> oldest
- We want "zoom out of card", but we probably shouldn't have "zoom-in to card" gesture until we figure out the interaction details around scroll snapping. I want to defer that until later.

paulrouget:
@gordonbrander file issues for these problems.

PR #480, opened 2015-07-14
Fix ZoomIn and ZoomOut

gordonbrander:
...by special-casing delegation of ZoomIn and ZoomOut actions


gordonbrander:
Fixes the zoom issues referenced in https://github.com/mozilla/browser.html/issues/476

Gozala:
Better fix for this is to instead wrap events in `WebView.Action({action: ZoomIn() })`

gordonbrander:
@Gozala I tried that first and it didn't work. Caused all webviews to disappear when action was sent. So there must be a bug in that code path somewhere.

PR #482, opened 2015-07-14
pinch should not show the awesomebar, and unpinch should show the webview

### LINKED ISSUE #469: No "zoom in" gesture
We want "zoom out of card", but we probably shouldn't have "zoom-in to card" gesture until we figure out the interaction details around scroll snapping. I want to defer that until later.

paulrouget:
Needs review.
fix #469


gordonbrander:
Yup

PR #487, opened 2015-07-14
Order preview cards LTR, new to old

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/468

### DIFF
--- a/src/browser/synthesis-ui.js
+++ b/src/browser/synthesis-ui.js
@@ -9,7 +9,6 @@ define((require, exports, module) => {
   const {Record, Any, Union} = require('common/typed');
   const {compose} = require('lang/functional');
   const WebView = require('./web-view');
-  const Preview = require('./web-preview');
   const Input = require('./web-input');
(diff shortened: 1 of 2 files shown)


src/browser/synthesis-ui.js
```diff
@@ -9,7 +9,6 @@ define((require, exports, module) => {
   const {Record, Any, Union} = require('common/typed');
   const {compose} = require('lang/functional');
   const WebView = require('./web-view');
-  const Preview = require('./web-preview');
```
gordonbrander: Remove unused variable

Gozala:
This works although it maybe a better idea to change `web-view.js` to use `unshift` vs `push` to just populate list in order we display it, specially since append to head is faster operation + we avoid reversing on every view.

Not a big deal though as with flex box is won't be an issue and may not matter in practice now either.

r+

gordonbrander:
@Gozala Here's a followup pull request using `unshift` as you suggest https://github.com/mozilla/browser.html/pull/497

PR #489, opened 2015-07-14
Fade in when dashboard -> webview

gordonbrander:
...and get rid of transition origin when scaling. If we want to set an origin on scaling, we have to be much smarter and take into account CSS centering as well as scroll position.

This is a temporary workaround for https://github.com/mozilla/browser.html/issues/488 that improves the situation a bit but does not solve the issue.


paulrouget:
Ok.

PR #490, opened 2015-07-14
WIP model modes

gordonbrander:
Attempting to model modes in a way that will give us enough information to fix https://github.com/mozilla/browser.html/issues/488.

### DIFF
--- a/src/browser/mode.js
+++ b/src/browser/mode.js
@@ -1,0 +1,45 @@
+define((require, exports, module) => {
+  const {Record, Union, Maybe} = require('common/typed');
+
+  // Define mode actions
+
+  const ShowWebView = Record({
+    // An annotation to describe how we activated the transition between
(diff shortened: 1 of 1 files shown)


src/browser/mode.js
```diff
@@ -0,0 +1,45 @@
...
+    description: 'Show create webview mode'
+  }, 'Mode.CreateWebView');
+  exports.CreateWebView = CreateWebView;
+
+  const Mode = Union({ShowWebView, EditWebView, CreateWebView});
+  exports.Mode = Mode;
+
+  // The mode model stores the current mode and previous mode.
+  const Model = Record({
+    curr: Mode,
```
gordonbrander: It looks like in other places, simple data is stored in the state. I'm not sure we should be storing the Mode Types in the state... However, I like the guarantees that the type system gives us if we do.

PR #491, opened 2015-07-14
Fix load stop.

### LINKED ISSUE #485: Stop doesn't work (pressing Escape while page is loading)

Gozala:
Fixes #485


src/browser/synthesis-ui.js
```diff
@@ -138,7 +139,13 @@ define((require, exports, module) => {
     state.mode === 'select-web-view' ? state.set('mode', 'show-web-view') :
     state;
 
+  const escape = state =>
+    state.made === 'show-web-view' ? state :
```
gordonbrander: Oops. I missed this. Should be `state.mode`.

Issue #484, opened 2015-07-14
When a webview is open in the background, it stays visible until it's loaded


paulrouget:
STR:
- open a web page
- Cmd-Click on a link
- a new webview is open
- stays above the current webview, disappear one is loaded

We used to keep the iframe `setVisible(true)` until the screenshot is done, but set the iframe `display:none` (or something like that). I guess now the `display:none` is not happening until the screenshot is done.

gordonbrander:
Fixed by https://github.com/mozilla/browser.html/pull/492

PR #492, opened 2015-07-15
Fix the open tab.

Gozala:
Fix for #484

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -260,6 +260,8 @@ define((require, exports, module) => {
   const updateByID = (state, id, action) =>
     action instanceof Load ? loadByID(state, id, action) :
     action instanceof Close ? closeByID(state, id) :
+    action instanceof Open ? open(state, action) :
+    action instanceof OpenInBackground ? open(state, action) :
     updateByIndex(state, indexByID(state, id), action);
   exports.updateByID = updateByID;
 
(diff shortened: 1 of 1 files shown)


src/browser/web-view.js
```diff
@@ -327,7 +329,7 @@ define((require, exports, module) => {
       display: 'none'
     },
     passive: {
-      visibility: 'collapsed'
+      visibility: 'collapse'
```
gordonbrander: Nit: `collapse` is an obscure property value meant for tables. We're taking advantage of the fact that it falls back to `hidden` on non-table elements. Better to use `hidden`. https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

Gozala: > Nit: collapse is an obscure property value meant for tables. We're taking advantage of the fact that it falls back to hidden on non-table elements. Better to use hidden. https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

The problem with `hidden` is that unlike `collapse` it still takes up a space even though it is not displayed.

gordonbrander: Hmm, this might be relying on a bug in our implementation. From https://developer.mozilla.org/en-US/docs/Web/CSS/visibility.

> For table rows, columns, column groups, and row groups the row(s) or column(s) are hidden and the space they would have occupied is removed (as if display: none were applied to the column/row of the table). __For other elements, collapse is treated the same as hidden.__

gordonbrander: If we want `display: none` we should probably use `display:none`.

gordonbrander: Filed a bug to follow up on this https://github.com/mozilla/browser.html/issues/495

PR #497, opened 2015-07-15
Model webviews LTR, newest to oldest

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/468

Do so by prepending models with `.unshift()` instead of `.push()`, per comment https://github.com/mozilla/browser.html/pull/487#issuecomment-121433663

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -206,13 +206,13 @@ define((require, exports, module) => {
 
   const open = (state, {uri, inBackground}) => state.merge({
     nextID: state.nextID + 1,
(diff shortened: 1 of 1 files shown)


paulrouget:
Can you make sure Ctrl-W, click on a link with a different target (like yahoo search) and Cmd-Tab select the right tab?

I tried `unshift` earlier, and I ran into issues. But I might have done something wrong.

gordonbrander:
@paulrouget `cmd tab`, `cmd shift tab` and `ctl w` on a link with a different target all work for me.

PR #499, opened 2015-07-15
WIP tailor mode transitions

gordonbrander:
WORK IN PROGRESS

- [x] Introduce mode `create-webview-quick`
- [x] Introduce mode `edit-webview-quick`
- [ ] Introduce mode `show-webview-quick`

This is take 2 on fixing https://github.com/mozilla/browser.html/issues/488. Trying a different approach in response to @Gozala's comments about take 1 https://github.com/mozilla/browser.html/pull/490.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -52,7 +52,7 @@ define((require, exports, module) => {
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
     'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.Action({action: WebView.Open()}),
+    'accel t': _ => WebView.Action({action: WebView.Open({source: 'keyboard'})}),
     'accel 0': _ => Shell.ResetZoom(),
     'accel -': _ => Shell.ZoomOut(),
     'accel =': _ => Shell.ZoomIn(),
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -51,7 +51,10 @@ define((require, exports, module) => {
 
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
-    'accel l': _ => Input.Action({action: Focusable.Focus()}),
+    'accel l': _ => Input.Action({
+      action: Focusable.Focus(),
+      source: 'keyboard'
+    }),
     'accel t': _ => WebView.Action({
       action: WebView.Open(),
       source: 'keyboard'
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -52,7 +52,10 @@ define((require, exports, module) => {
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
     'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.Action({action: WebView.Open({source: 'keyboard'})}),
+    'accel t': _ => WebView.Action({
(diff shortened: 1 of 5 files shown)


gordonbrander:
Idea: perhaps we should be describing transition states as modes, too. So a transition from creating a page on the dashboard to showing the page would produce the following mode switches:

    create-webview -> transition-from-create-webview-to-show-webview-quick -> show-webview

This would let us be more specific about how we render things during the transition -- something we currently have a problem with. For example, we could render the dashboard during the transition so that the webview is crossfaded above it. Then when we have fully transitioned to `show-webview` we could stop rendering the dashboard.

src/browser/browser.js
```diff
@@ -51,8 +51,14 @@ define((require, exports, module) => {
 
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
-    'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.Action({action: WebView.Open()}),
+    'accel l': _ => Input.Action({
+      action: Focusable.Focus(),
+      source: 'keyboard'
```
gordonbrander: To communicate this information, I'm unsure if we should annotate the action or create a new action type like `Focusable.KeyboardFocus` or `WebView.KeyboardOpen`.

src/browser/web-preview.js
```diff
@@ -275,14 +275,17 @@ define((require, exports, module) => {
     ]);
 
   const viewInEditMode = (loaders, pages, selected, theme, address) =>
-    viewContainer(theme, ...viewPreviews(loaders, pages, selected, address));
+    viewContainer(theme, ghostPreview, ...viewPreviews(loaders, pages, selected, address));
```
gordonbrander: Always showing ghost preview to avoid issues with "jumping" caused by rendering edit mode when transitioning from `create-web-view -> show-web-view`.

gordonbrander:
The logic in `synthesis-ui.js` is getting increasingly knotty.

Gozala:
I'm afraid that encoding too much informarion into state will complicate reasoning about the flow. If I need to know **from** which state I arrived to **current** state and by which **animation** to make call about rendering that is three factor complexity. What I was suggesting `update` should be able to look at the **current** `state` and `action` To make a call about what the next state will be. If it happens to be that prior **A** state prevents current **B** state from transitioning to next **C** state then **A** should not be able to transition to **B** state in first place instead it should have transitioned to **B2** state. In other words state machine (& its branching shape) should be a tree not a graph.

In case of `view` it seems that we need not only state but also an info about `animation` we'd like to view while getting there. It seems that it would make sense to have `anation` field alog with `mode` just for that. 

As of state changes during animation it seems that we need either:

1. intermidiate states to prevent state machine from jumping a state till transition is finished.
2. Clear animation field once transition is done and prevent state changes while transitioning

But honestly blocking interactions during transitions seems odd & likely will be confusing and frustrating for end user. I would much rather think of UX where such user actions can be handled nicely. As I mentioned before I think persistant new card does in fact make a simpler & more intuitive interaction model something that I really liked from previous iteration.

Gozala:
Btw currently focusing input or page causes mode changes but that does not neceserily has to be a case and it may make sense to change if that would solve some of the issues

gordonbrander:
Fine with all of this, but do want to note that the ghost card issue is orthogonal. I think it happens to be the right choice from a UX perspective, but we still may need to distinguish between `edit`, `switch` and `create` states... for example, card may appear with with different animations.

gordonbrander:
Ok, trying a 3rd approach: https://github.com/mozilla/browser.html/pull/509

PR #501, opened 2015-07-16
Update .gitignore

GabrielDuque:


Gozala:
Can you clarify why should we add `.LOGS-ID` ? I do not believe we generate it.

Gozala:
Given your other pulls I decided to close this one.

PR #507, opened 2015-07-20
Fix color themes

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/467. Fix hardcoded color themes and themes generated using `theme-color`.

Will eventually follow up with a PR that bases themes on `background-color` instead of `theme-color` meta tag.


Gozala:
R+

PR #509, opened 2015-07-21
Tailor mode transitions (take 2)

gordonbrander:
Fix for https://github.com/mozilla/browser.html/issues/488.

In this version, we have a `mode` field and a `transition` field in the model.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -52,7 +52,7 @@ define((require, exports, module) => {
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
     'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.TransitionToOpenWithFade(),
+    'accel t': _ => WebView.FadeToOpen(),
     'accel 0': _ => Shell.ResetZoom(),
     'accel -': _ => Shell.ZoomOut(),
     'accel =': _ => Shell.ZoomIn(),
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -37,7 +37,7 @@ define((require, exports, module) => {
   const Model = Record({
     version: '0.0.7',
     mode: 'create-web-view', // or show-web-view, edit-web-view, choose-web-view
-    transition: 'normal', // or quick
+    transition: 'zoom', // or fade
     shell: Focusable.Model({isFocused: true}),
     updates: Updates.Model,
     webViews: WebView.Model,
@@ -51,14 +51,8 @@ define((require, exports, module) => {
 
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
-    'accel l': _ => Input.Action({
-      action: Focusable.Focus(),
-      source: 'keyboard'
-    }),
-    'accel t': _ => WebView.Action({
-      action: WebView.Open(),
-      source: 'keyboard'
-    }),
+    'accel l': _ => Input.Action({action: Focusable.Focus()}),
+    'accel t': _ => WebView.TransitionToOpenWithFade(),
     'accel 0': _ => Shell.ResetZoom(),
     'accel -': _ => Shell.ZoomOut(),
     'accel =': _ => Shell.ZoomIn(),
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -37,7 +37,7 @@ define((require, exports, module) => {
   const Model = Record({
     version: '0.0.7',
     mode: 'create-web-view', // or show-web-view, edit-web-view, choose-web-view
-    transition: 'zoom', // or fade
+    transition: 'zoom', // or fade, none
     shell: Focusable.Model({isFocused: true}),
     updates: Updates.Model,
     webViews: WebView.Model,
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -7,7 +7,7 @@ define((require, exports, module) => {
   'use strict';
 
   const {html, node, render, cache} = require('reflex');
-  const {Record, Any, Union} = require('common/typed');
+  const {Record, Any, Union, Maybe} = require('common/typed');
   const {inspect} = require('common/debug');
   const {StyleSheet, Style} = require('common/style');
   const WindowBar = require('./window-bar');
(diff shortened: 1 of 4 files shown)


src/browser/synthesis-ui.js
```diff
@@ -120,16 +134,24 @@ define((require, exports, module) => {
...
+    (action instanceof WebView.Open && !action.uri) ?
+      createWebView(state, 'normal') :
+    action instanceof WebView.SelectNext ?
+      selectNext(state) :
+    action instanceof WebView.SelectPrevious ?
+      selectPrevious(state) :
     state;
 
-  const updateByInputAction = (state, action) =>
+  const updateByInputAction = (state, source, action) =>
```
Gozala: `source` argument here seems obsolete.

Gozala: It looks like it is actually used so ignore my previous comment.

src/browser/browser.js
```diff
@@ -50,8 +51,14 @@ define((require, exports, module) => {
 
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
-    'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.Action({action: WebView.Open()}),
+    'accel l': _ => Input.Action({
+      action: Focusable.Focus(),
+      source: 'keyboard'
```
Gozala: Not entirely sure where the `source` field was introduced but I'd recommend against it as with it we end up with multi-level polymorphism. What I mean by that is we use types to annotate a raw data, if we introduce `source` field it kind of serves same purpose but in a different style which means we'll have two types of annotations to check to decide how to act. What I propose to de instead is define a different type for that for example `Input.KeyboardAction({action})` that way you'll be able to distinguish between actions triggered by keyboard or differently by checking `instanceof`.

I don't know enough context here to say for sure but in general I'd recommend to avoid distinction by a triggering source and instead use of natural terms for naming actions. For example we could have `EditLocation` action & `TransitionToLocationEdit` action that would transition state to the same mode but with a different animation.

I think that is more inline with a desired effect because `source: 'keyboard'` does not necessary have to translate to quick / slow animation. Hypothetically we make have another keyboard shortcut say `accept shift t` that  behaves identically to clicking **+** button in which case encoding source of an action won't be able to handle it.

gordonbrander: I see. I wondered about this and left a request-for-comment on the prev PR. Glad to have clarification. I'll update the approach here.

gordonbrander:
We could get fancier and rename `WebView.Open` to `TransitionToOpenWithZoom`, replace `Focusable.Focus` with `TransitionToEditWithFade` etc, but I wanted to take the minimum viable approach first.

src/browser/browser.js
```diff
@@ -37,6 +37,7 @@ define((require, exports, module) => {
   const Model = Record({
     version: '0.0.7',
     mode: 'create-web-view', // or show-web-view, edit-web-view, choose-web-view
+    transition: 'zoom', // or fade, none
```
Gozala: Maybe it's better to use `null` instead of  `'none'` ? I don't feel strongly on either option, just a thought I had when I saw `'none'`.

src/browser/browser.js
```diff
@@ -51,7 +52,7 @@ define((require, exports, module) => {
   const modifier = OS.platform() == 'linux' ? 'alt' : 'accel';
   const KeyDown = KeyBindings({
     'accel l': _ => Input.Action({action: Focusable.Focus()}),
-    'accel t': _ => WebView.Action({action: WebView.Open()}),
+    'accel t': _ => WebView.TransitionToOpenWithFade(),
```
Gozala: Would not `FadeToOpen` be descriptive enough ?

src/browser/synthesis-ui.js
```diff
@@ -64,10 +76,12 @@ define((require, exports, module) => {
                        state.getIn(['webViews', 'loader', index, 'uri']));
   };
 
+  const editWebView = switchMode('edit-web-view', 'fade');
```
Gozala: Is it always with fade ? Otherwise name `editWebView` is not good enough.

src/browser/synthesis-ui.js
```diff
@@ -121,12 +135,20 @@ define((require, exports, module) => {
...
-    (action instanceof WebView.Open && !action.uri) ? createWebView(state) :
-    action instanceof WebView.SelectNext ? selectNext(state) :
-    action instanceof WebView.SelectPrevious ? selectPrevious(state) :
+    action instanceof Focusable.Focus ?
+      showWebViewByID(state, id) :
+    action instanceof Focusable.Focused ?
+      showWebViewByID(state, id) :
+    action instanceof WebView.Close ?
+      closeWebViewByID(state, id) :
+    (action instanceof WebView.Open && !action.uri) ?
```
Gozala: Sidenote (to myself ?): Maybe jamming `createWebView` onto `WebView.Open` was not a good idea as `WebView.Open` seem to do totally different things when `uri` is present. Maybe we should have `WebView.Open` and `SynthesisUI.CreateWebView` or something.

Gozala: P.S.: I'm not entirely sure that `Open({uri: null})` even happens now that `TransitionToOpenWithFade` is used. If it is not maybe it makes sense to make `uri` mandatory on `Open`.

src/browser/synthesis-ui.js
```diff
@@ -49,14 +49,26 @@ define((require, exports, module) => {
...
     state.set('webViews', WebView.selectByID(state.webViews, id));
 
-  const showWebView = switchMode('show-web-view');
-  const showWebViewByID = compose(showWebView, selectViewByID);
+  const showWebViewByID = compose(
+    switchMode('show-web-view', 'zoom'),
+    selectViewByID
+  );
+
+  const showWebViewByIDFade = compose(
```
Gozala: Nit: I'd call it `fadeIntoWebViewByID`

src/browser/web-view.js
```diff
@@ -393,23 +398,49 @@ define((require, exports, module) => {
...
       transform: 'scale(0)',
       opacity: 0,
       pointerEvents: 'none',
     }
   });
 
-  const view = (mode, loader, shell, page, address, selected) => {
-    const isActive = mode === 'show-web-view';
+  const view = (mode, transition, loader, shell, page, address, selected) => {
+    const additionalStyles =
```
Gozala: Can we have something like `Style(webviewsStyle.base, webviewsStyle[getModeStyle(mode, transition)])` instead of inlining all this logic.

Gozala:
r+ just made some nit comments

src/browser/synthesis-ui.js
gordonbrander: @Gozala I wondered about this too. Seems like a good idea to clean this up in a follow-up PR.

gordonbrander:
Addressed all nits and even found and fixed a small bug.

PR #514, opened 2015-07-23
show url in awesomebar instead of title for history entries

paulrouget:


src/browser/suggestion-box.js
```diff
@@ -202,7 +204,7 @@ define((require, exports, module) => {
       }, Icon[type] || ''),
       html.span({
         key: 'suggestion'
-      }, state.title)
```
gordonbrander: We should update this to match the spec, which displays `{title} — {domain}` for history entries.

src/browser/suggestion-box.js
```diff
@@ -202,7 +204,7 @@ define((require, exports, module) => {
...
+      }, type == 'history' ? state.uri : state.title)
```
gordonbrander: We should update this to match the spec, which displays `{title} — {domain}` for history entries.

PR #515, opened 2015-07-23
make loader uri optional

paulrouget:
It's an unnecessary constrain. If I want to open empty pages for example, this makes it impossible.

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -256,7 +256,7 @@ define((require, exports, module) => {
     const loader = state.loader.get(index);
     return !loader ?
             open(state, action) :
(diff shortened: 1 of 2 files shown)


src/browser/web-view.js
```diff
@@ -256,7 +256,7 @@ define((require, exports, module) => {
     const loader = state.loader.get(index);
     return !loader ?
             open(state, action) :
-           URI.getOrigin(loader.uri) !== URI.getOrigin(action.uri) ?
+           loader.uri && (URI.getOrigin(loader.uri) !== URI.getOrigin(action.uri)) ?
```
gordonbrander: Why do we do this?

paulrouget: Because uri is now optional, and getOrigin will fail if uri is undefined.

PR #516, opened 2015-07-23
allow looping through pages by default

paulrouget:


gordonbrander:
:+1:. We'll have to revisit this when we start managing large numbers of cards. When you have a large number, we should probably animate over when looping.

PR #517, opened 2015-07-23
do not update overflow once set to true

paulrouget:


src/browser/web-page.js
```diff
@@ -113,7 +113,7 @@ define((require, exports, module) => {
   const update = (state, action) =>
     action instanceof TitleChanged ? state.set('title', action.title) :
     action instanceof IconChanged ? state.set('icon', action.icon) :
-    action instanceof OverflowChanged ? state.set('overflow', action.overflow) :
+    action instanceof OverflowChanged ? state.set('overflow', state.overflow || action.overflow) : // we don't want overflow to be set back to false
```
gordonbrander: I think this would be better written:

    action instanceof OverflowChanged && !state.overflow ? state.set('overflow', action.overflow)

But I could be wrong. cc @Gozala.

paulrouget: Ok.

PR #519, opened 2015-07-23
Favicon selection mechanism

### LINKED ISSUE #455: favicons aren't always displayed, even when they exist
Looks like we respond to `onIconChange: Event('mozbrowsericonchange')`, but I suspect this doesn't fire unless the page has a `<link rel="shortcut icon">`.

cc @Gozala

paulrouget:
Fixes #455

### DIFF
--- a/src/browser/web-page.js
+++ b/src/browser/web-page.js
@@ -11,17 +11,25 @@ define((require, exports, module) => {
   const Loader = require('./web-loader');
   const Progress = require('./web-progress');
   const WebView = require('./web-view');
+  const Favicon = require('common/favicon');
 
   // Model
 
 
   const Heros = List(String, 'Heros');
 
+  const Icon = Record({
+    href: String,
+    sizes: Maybe(String),
+    rel: Maybe(String),
+  });
+
   const Model = Record({
     title: Maybe(String),
 
     label: Maybe(String),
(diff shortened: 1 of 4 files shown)


src/browser/web-page.js
```diff
@@ -110,9 +118,15 @@ define((require, exports, module) => {
       hero: action.hero
     });
 
+  const updateIcon = (state, icon) => {
+    const {bestIcon, faviconURL} = Favicon.getBestIcon([state.icon, icon]);
+    console.log(faviconURL);
```
gordonbrander: Plz remove `console.log(faviconURL)`.

src/browser/web-view.js
```diff
@@ -518,7 +518,7 @@ define((require, exports, module) => {
   Event.mozbrowsertitlechange = ({target, detail: title}) =>
     TitleChanged({uri: target.location, title});
 
-  Event.mozbrowsericonchange = ({target, detail: {href: icon}}) =>
+  Event.mozbrowsericonchange = ({target, detail: icon}) =>
```
gordonbrander: Should we change the field name from `detail` to `href` or `uri`?

paulrouget: `detail` is an event field. Not a thing we can rename.

PR #520, opened 2015-07-23
fix cmd-click

paulrouget:
Cmd-Click doesn't open links in background anymore. Regressed because of #497.


gordonbrander:
Looks good. We should follow up later with an "open in background" animation https://github.com/mozilla/browser.html/issues/524.

PR #521, opened 2015-07-23
select younger tab on close

paulrouget:
Other wise, pressing Cmd-W brings you to the older tab, which is weird.


gordonbrander:
:+1:

PR #528, opened 2015-07-29
Basic browser

paulrouget:
This is an implementation of a basic browser.

No need to review the content of `/browser2/`.


paulrouget:
I suggest we use a different repository (mine) to host this code.

Issue #513, opened 2015-07-23
awesomebar is white on white


paulrouget:
The awesomebar is all white.

It's because it uses `theme.foreground` which doesn't exist, so the color property is inherited from the shell, which is white.

paulrouget:
@gordonbrander can you please look at this?

PR #535, opened 2015-07-30
(WIP) Fix location bar styling when value is ''

gordonbrander:
This is a follow-up to #513... a case I missed where location bar is focused, but you haven't typed anything yet.


paulrouget:
Isn't testing `input.value !== ''` supposed to be the same as testing `input.value`?

gordonbrander:
@paulrouget you're right. I copied this approach from the other place we test for whether we're showing suggestions. However, I found yet another case that it doesn't cover. I'll push an update to this branch that introduces an `isSuggesting` predicate function.

gordonbrander:
Going to squash the commits.

PR #541, opened 2015-08-03
Update path from B2G to Browser.html

Gozala:


Gozala:
@paulrouget or @gordonbrander I think this was missed during rename.

PR #544, opened 2015-08-03
Fix location bar styling when value is ''

gordonbrander:
Follow-up for https://github.com/mozilla/browser.html/issues/513... a case I missed where location bar is focused, but you haven't typed anything yet.

Fixes https://github.com/mozilla/browser.html/issues/539.
Fixes https://github.com/mozilla/browser.html/issues/537.
Fixes https://github.com/mozilla/browser.html/issues/540.
Fixes https://github.com/mozilla/browser.html/issues/545.

### DIFF
--- a/src/browser/location-bar.js
+++ b/src/browser/location-bar.js
@@ -207,7 +207,6 @@ define((require, exports, module) => {
     // Make forwarding addres that wraps actions into `Input.Action`.
     const inputAddress = address.forward(InputAction);
 
-    // If we have suggestions and the input isn't empty
     const view = Suggestions.isSuggesting(input, suggestions) ?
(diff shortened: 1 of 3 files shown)


src/browser/location-bar.js
```diff
@@ -208,7 +208,7 @@ define((require, exports, module) => {
     const inputAddress = address.forward(InputAction);
 
     // If we have suggestions and the input isn't empty
-    const view = (input.value && input.isFocused) ?
+    const view = Suggestions.isSuggesting(input, suggestions) ?
```
Gozala: Nit: Probably comment above needs update or can be removed entirely.

Gozala:
Looks good to me r+

Issue #527, opened 2015-07-29
Prerender app before packaging


paulrouget:
We need to render the app on load. We can't wait for react and all the modules to load.

PR #523 introduces a way to save previous rendering into localstorage. But this won't work for the first load. We can fill the `<body>` of `index.html` when we package the app.  React allows that (server side rendering).

`npm run package` should run a node script that generate the HTML code, and append it to `index.html` or `prerendering.js`.

Gozala:
Closing this for now as #547 covers it, any further improvements can be tracked separately.

Issue #522, opened 2015-07-23
Package JS code with r.js, babel and uglify


paulrouget:
We want to package browser.html.

We can't use r.js out of the box for 2 reasons:
- `require('common/worker!...')` is not supported
- uglifyjs doesn't support ES6

I'm not comfortable with all these tools, but here is what I managed to do so far:

- use react.min
- update requirejs to latest version (necessary)
- comment out worker code
- `build.js`: https://gist.github.com/paulrouget/0ba08478ae7a7e6d2299
- /node_modules/requirejs/bin/r.js -o ./build.js
- it creates a file named `main-built.js`
- `babel main-built.js > main-built-es5.js`
- I had to add a call to `super()` in element.js (see https://github.com/babel/babel/issues/1131). ES6 code doesn't work anymore, but ES5 code works
- `./node_modules/uglify/bin/uglify -s ./main-built-es5.js -o ./main-built-es5-min.js`

I end up with a file that is ~280K, which is not bad (600k before).

@Gozala can you take over? You'll do a better job than me.

paulrouget:
First step comes with PR #526.

Gozala:
Done in #547

PR #547, opened 2015-08-05
Packaging.

Gozala:
This uses babel and bunch of other tools to generate minimized js bundles:

- browser/index
- service/history-worker
- about/settings/index


Gozala:
@paulrouget can you please take a loot at this ? 
This makes load & render sync but I'm afraid initial one takes too long.

Gozala:
This addresses #522 and #527 (to some degree) as with this js load and initial render becomes synchronous.

gulpfile.js
```diff
@@ -0,0 +1,121 @@
...
+var path = require('path');
+
+var settings = {
+  cache: {},
+  packageCache: {},
+  debug: true,
+  watch: false,
+  compression: null
+};
+
```
paulrouget: Are these settings good for a release build?

paulrouget:
This looks good, but indeed, this slows down the first rendering quite a bit.

Let's keep it that way for now. We will figure prerendering in a future PR.

Feel free to merge.

gulpfile.js
Gozala: > Are these settings good for a release build?

Yes, but that's an excellent question, I'll add comments to explain what those prefs are for.

Issue #486, opened 2015-07-14
Session restore breaks often


gordonbrander:
This seems to happen most often when refreshing on the dashboard:

- Load page
- Navigate to card dashboard
- Hit `fn f5`
- Session is not restored!

...but not always. Sometimes we fail to restore even when viewing a page.

See also https://github.com/mozilla/browser.html/issues/476#issuecomment-121112262

Gozala:
I have figured out the issue which is relate to selection state, when we select all we set selection to `{start: 0, end: Infinity}` but `JSON.serialize({start: 0, end: Infinity})` produces `{start:0, end:null}` which causes invalid state to be stored for the session.

The for this issue is one liner that I've included in the different pull:
https://github.com/mozilla/browser.html/pull/547/files#diff-2c8b4521eac83629c3dc939a131e56f7R229

I can cherry-pick it to land separately.

Gozala:
@gordonbrander wanna look at the pull and see if resolves the issue, also note that this was regression introduced during refactoring that also breaks test selection in the input which is also fixed by this change. Basically when we update selection by setting it to `{start: 0, end: Infinity}` that causes `selection` event that should contain actual integer `end` value that is then stored into state there for we avoid bug caused by `JSON.serialize` outlined above.

PR #548, opened 2015-08-05
Fix session restore bug

Gozala:
Fix for #486

So issue is related to selection state, which is set to `{start: 0, end: Infinity}` when we select all of the input. Although if such app state is stored it causes `JSON.serialize({start: 0, end: Infinity})` to run that produces `'{start:0, end:null}'` which is invalid state, given that `selection.end` is declared to be a number not a `null`.

Solution is to react to `selection` events and update state when the happen (please note that this was a case prior to refactoring, the fact that it is no longer a case was unintentional). Now with this change when we update selection by setting it to `{start: 0, end: Infinity}` that causes a `selection` event that will contain integer value for `selection.end` which is then stored into state, which in turn avoids an outlined bug caused by `JSON.serialize`.

### DIFF
--- a/src/browser/location-bar.js
+++ b/src/browser/location-bar.js
@@ -228,6 +228,7 @@ define((require, exports, module) => {
         isFocused: input.isFocused,
         selection: input.selection,
         onChange: inputAddress.pass(Change),
+        onSelect: inputAddress.pass(Change),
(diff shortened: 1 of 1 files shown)


paulrouget:
thanks.

PR #549, opened 2015-08-05
Properly handle selection change.

Gozala:
I'm afraid I have introduced regression in #548 that breaks suggestion-box as it sends change events whenever selection changes. I think suggestion engine needs a deeper changes, but this does solve an issue right now so it's worth doing it and rethink how we deal with suggestion later on.


Gozala:
@gordonbrander can you please review this ?

gordonbrander:
Works!

PR #550, opened 2015-08-05
Devtools shortcut

Gozala:
`F12` is taken by OSX by default, there for I'm adding another keybinding that matches regular devtools key binding on all browsers.


paulrouget:
Works for me.

PR #552, opened 2015-08-06
Fix scraper to make it use promise properly.

Gozala:
Babel replaces all uses of `Promise` to `require('babel-core/runtime').Promise` that is either native promise or polyfill. This introduced a regression as we use `scraper` function with `Promise(...)` as `iframe.executeScript(scraper + '()')` which now breaks as `Promise` there will be replaced with `require('babel-core/runtime').Promise` and in the context it's run there is no `require`. Fix simply changes use to `window.Promise` so babel won't replace it and there for we won't run into this issue.

### DIFF
--- a/src/service/scraper.js
+++ b/src/service/scraper.js
@@ -352,7 +352,7 @@
 
     // @TODO need some methods for scaling and cropping images.
 
-    const ready = new Promise((resolve) => {
+    const ready = new window.Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


gordonbrander:
:+1:

As a follow-up, any way we can turn off Babel's Promise polyfill?

gordonbrander:
Looks like `babel/polyfill` is all or nothing. This is the right approach.

Issue #536, opened 2015-07-30
Navigating from page to page with color theme results in flashes of color


gordonbrander:
1. Open http://imgur.com
2. Click on an image.

You'll see the header fades from `#222 -> #FFF -> #222` quickly.

gordonbrander:
The question is what to do when we have a previous page with a pallet. Carry it over? This would mean that clicking on a link to a different domain could carry the previous pallet with you. That would be a bug

Really what we want to do is load opened pages in the background for a moment until we're ready to switch.

cc @Gozala: could you suggest the best approach here?

KevinGrandon:
We had a similar problem in FxOS. You do get the metachange event before the page is loaded as typically the meta tag is included in the initial HTML payload. I believe we waited until loadend (instead of locationchange) and if no metachange was found, then we display the default.

paulrouget:
> Really what we want to do is load opened pages in the background for a moment until we're ready to switch.

Define "ready to switch". We show the page as soon as it starts loading. Delaying would be against the spec and will most likely break pages.

I think we should reset the pallet after a short delay. Like 2 seconds.

gordonbrander:
@paulrouget I'm pretty sure this is how Firefox and other browsers behave today:

1. Click link
2. Start loading immediately
3. Show page on first paint, or after ~2.5s delay.

This does a few things:

- You can "change your mind" and click a different link, interrupting the first.
- By skipping the "flash of white" you see fewer "noise frames", which makes things feel faster.

If we did something similar, we would be able to theme the location bar consistently, since this information would be available by that time.

paulrouget:
This is how it works:

1. Click link
2. Connecting to the server (no response yet), page (1) stays up
3. first response from the server, page starts loading, page (1) is trashed
4. first paint of the page

Use can "change its mind" during (2).

Both Firefox and Graphene work the same way.

But during phase 2 we trash the pallet. We should trash it only at (3) (when we get the "loading" event).

gordonbrander:
Ok, that sounds good.

paulrouget:
iirc, we get the loading event at connection. I think we will need to introduce more granular events. We need to re-land bug 1125713.

Gozala:
So right now what happens is that on `LoadStrated` we reset the page state which includes thumbnail, scraper data and also a color pallet:

https://github.com/mozilla/browser.html/blob/50307661782181eff31a165b7eab63bfcc38fce3/src/browser/web-page.js#L134

[...]

PR #554, opened 2015-08-06
Keep the color pallet across domains.

Gozala:
This is imperfect fix for #536 that resets state of the page card only on location change if the location has changed, but keeps the color pallet unless domain has changed.

### DIFF
--- a/src/browser/iframe.js
+++ b/src/browser/iframe.js
@@ -47,6 +47,8 @@
     onUserActivityDone: Event('mozbrowseractivitydone'),
     onVisibilityChanged: Event('mozbrowservisibilitychange'),
     onMetaChanged: Event('mozbrowsermetachange'),
+    onFirstPaint: Event('mozbrowserfirstpaint'),
+    onDocumentFirstPaint: Event('mozbrowserdocumentfirstpaint'),
     // Use `VirtualEvent` to proxy events in order to mutate `target.location`
     // so that user can check `target.location` before deciding if change to
     // `target.src` is required.
--- a/src/browser/web-page.js
+++ b/src/browser/web-page.js
@@ -9,6 +9,7 @@
   const Progress = require('./web-progress');
   const WebView = require('./web-view');
   const Favicon = require('../common/favicon');
+  const URI = require('../common/url-helper');
 
   // Model
 
@@ -22,6 +23,7 @@
   });
 
   const Model = Record({
+    uri: Maybe(String),
     title: Maybe(String),
 
     label: Maybe(String),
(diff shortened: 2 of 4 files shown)


KevinGrandon:
How come you don't want to keep track of when the theme-color changed? You would do this by checking if you got a metachange event between locationchange and loadend.

Gozala:
> How come you don't want to keep track of when the theme-color changed? You would do this by checking if you got a metachange event between locationchange and loadend.

I gave it another try and I think this addresses that.

KevinGrandon:
> I gave it another try and I think this addresses that.

The only problem is if two pages on the same origin uses a different theme color. That's possible, but I agree that it's pretty unlikely.

Gozala:
> The only problem is if two pages on the same origin uses a different theme color. That's possible, but I agree that it's pretty unlikely.

This implementation should cover that though, I no longer care about the domain just expect that metachange will happen between loadstart and loadend and if it did not I reset pallet. So if you stay on the same domain but get metachange event colors will be updated as expected.

KevinGrandon:
Ok, after reading the code again it looks more clear now to me. R+, but I don't really know your code too well so if you want another review that's totally fine by me :)

gordonbrander:
Header displays previous pallet when new page from different domain is shown. The pallet is replaced by the time page load completes, but this usually takes a few seconds.

screen-shot-2015-08-06-at-10 51 24-

Is this expected?

gordonbrander:
Looks like we also have some timing issues:

screen shot 2015-08-06 at 11 02 03

Gozala:
@gordonbrander I described the logic it follows here:

https://github.com/mozilla/browser.html/issues/536#issuecomment-128201073

> Made a second pass on this, I think this is more complete solution. Now color pallet is changed either by lookup in curated list of domains (takes a precedence) or by a meta tag containing theme-color if pallet has not being updated by either source between loadstart and loadend then pallet is reset. This eliminates intermediate color changes during navigation.

So yes that is expected behavior given current implementation, but maybe there is a better way to go about it I don't know. I think no matter how we approach this we would either have out of date colors for some time frame or we will have a flicker in some cases.

Gozala:
Ok did yet another iteration on this, now logic is following:

- `loadstart`: clear extracted colors (via meta change or curated list) for the page, but keep the current pallet
- `locationchange`: set a `curatedColor` for a page if domain is curated
- `metachange` (may not occur): set a `themeColor` for a page
- `documentfirstpaint`: (happens on every load from what I can tell, but not on goBack): generate pallet out of `curatedColor` or `themeColor` or use default if non of the colors available.
- loadend: repeat what we do in `documentfirstpaint` to cover goBack / goForward case.

gordonbrander:
Still some delay, but that's probably as good as we can do unless we have better `paint` events.

paulrouget:
> Still some delay, but that's probably as good as we can do unless we have better paint events.

@gordonbrander Can you explain?

PR #555, opened 2015-08-06
Fix issue with window buttons.

Gozala:


gordonbrander:
:+1: `<base href>` brings nothing but pain.

PR #561, opened 2015-08-06
Add scroll snapping

gordonbrander:
This adds scroll snapping to tab cards. Fixes https://github.com/mozilla/browser.html/issues/558.

@paulrouget The physics don't feel quite right on Desktop. Let's see if we can tweak the platform.


Gozala:
R+

Issue #568, opened 2015-08-07
setVisible on `event.detail.frameElement` throws


Gozala:
Attempt to use `frame.setVisible(value)` on `event.detail.frameElement` from `mozbrowseropen` event throws:

`InvalidNodeTypeError: The supplied node is incorrect or has an incorrect ancestor for this operation.`

gordonbrander:
@Gozala should this be open still?

Gozala:
Yes it should be fixed in gecko.

Gozala:
Our target runtime is on Servo so no point in keeping gecko issues open.

Issue #567, opened 2015-08-07
zoom on `event.detail.frameElement` throws


Gozala:
Attempt to use `frame.zoom(n)` on `event.detail.frameElement` from `mozbrowseropen` event throws:

 `InvalidNodeTypeError: The supplied node is incorrect or has an incorrect ancestor for this operation.`

gordonbrander:
@Gozala should this be closed?

Gozala:
No it should be fixed in gecko.

Gozala:
Our target runtime is on Servo so no point in keeping gecko issues open.

Issue #566, opened 2015-08-07
Pass config handle to mozbrowseropen event instead of frameElement 


Gozala:
Currently `window.open()` causes `mozbrowseropen` event on an iframe with `event.detial.frameElement` that can't be used after animation frame (see #565). Attempt to workaround that limitation by stashing it to `document.head` and then reparenting on animation frame into appropriate location causes a crash (see #564).

Truth is this API is not only buggy but also really unfortunate as it does not play well with a popular libraries like react and is quite limiting as it is. As far as I can tell only reason to pass an iframe element with an event is to attach a specific configuration to it so that opener can remain connected to it. It would be much better if runtime stored that configuration / internal iframe implementation internally and exposed to a user a unique id for it that could be used by an embedder to pass that configuration to an iframe by assigning attribute with that id. In fact Electron seems to already do exactly that by passing `event.frameName` (see http://electron.atom.io/docs/v0.30.0/api/web-view-tag/#new-window)

paulrouget:
I might be completely wrong because I don't know Electron, but I believe they have a very basic implementation of `window.open` *from a child process*. It basically works the same way we used the `mozbrowseropenwindow` event. You take the URL, create a new iframe, and that's it. There is no inter window communication system and there is no process management.

`frameElement` and `frameName` are not equivalent, as `frameElement` points to an existing window, whereas `frameName` points to nothing, it's just the name used in `window.open()` and maybe the `target`, and we actually support that in Gecko already: `event.detail.name`.

If I'm wrong (and again, I might really be, I haven't look too much), I'd be happy to look at how it works internally in Electron/Webkit.

paulrouget:
Let's forget about the sync issue (discussed in #565) and let's focus on the id-over-frame proposal.

Gaia relies on this API, and it seems to work well for them.
What's special about us that makes it difficult to use?

gordonbrander:
@Gozala should this be open still?

Gozala:
@gordonbrander I think it should at least for now.

Gozala:
> Let's forget about the sync issue (discussed in #565) and let's focus on the id-over-frame proposal.

Well it's hard to discuss the issues if we ignore some of them.


> Gaia relies on this API, and it seems to work well for them.

That does not mean that API is sane.

> What's special about us that makes it difficult to use?

[...]

Issue #564, opened 2015-08-07
reparenting `event.detail.frameElement` after animation frame causes a crash


Gozala:
Blocks  #483

 I'll try to create a simple test case and then submit a b2g bug to bugzilla

Gozala:
Ok crash seems to happen if `iframe` is inserted into `head` and only afterwards under desired parent. Stashing it into `document.body` seems to not cause a crash, but reparenting also seems to cause `frameElement` to loose it's special connection with an opener.

gordonbrander:
@Gozala should this still be open?

Gozala:
I think this is a bug that is worth fixing on gecko side, that being said I don't really care if we keep it open or not.

Gozala:
No need to keep this open on our side as we can't really do anything about it.

PR #569, opened 2015-08-07
Shameful fix for #475 that makes me feel guilty.

Gozala:
This is very hacky fix that has to workaround ton of issues, here is little overview:

- Workaround #566 and present desired API by boxing frame in a string instance that is later unboxed by an IFrame react component.
- Workaround #567 by setting a flag on iframe so that implementation of `IFrame` won't call `zoom` on it.
- Workaround #568  by setting a flag on iframe so that implementation of `IFrame` won't call `setVisible` on it. 
- Workaround #564 by creating a type `Force.Action` so that actions boxed with it will force render loop to run sync in the same tick.

### DIFF
--- a/src/browser/iframe.js
+++ b/src/browser/iframe.js
@@ -4,9 +4,54 @@
   'use strict';
 
   const Focusable = require('../common/focusable');
-  const {Element, BeforeAppendAttribute, VirtualAttribute, Event, VirtualEvent} = require('../common/element');
+  const {ElementView, BeforeAppendAttribute, VirtualAttribute, Event, VirtualEvent} = require('../common/element');
+  const React = require('react');
+  const {html} = require('reflex');
 
-  const view = Element('iframe', {
+  /*
+  const DOMProperty = require('react/lib/ReactInjection').DOMProperty;
+
+  // Configure react to make it understand custom mozbrowser attributes.
+  DOMProperty.injectDOMPropertyConfig({
+    Properties: {
+      'remote': DOMProperty.MUST_USE_ATTRIBUTE,
+      'mozbrowser': DOMProperty.MUST_USE_ATTRIBUTE,
+      'mozapp': DOMProperty.MUST_USE_ATTRIBUTE,
+      'mozallowfullscreen': DOMProperty.MUST_USE_ATTRIBUTE,
+    }
+  });
+  */
+
+
+  class IFrameView extends ElementView {
+    componentDidMount() {
+      super.componentDidMount();
+      this.setState({swapped: true});
+    }
+    render() {
+      return React.createElement(this.props.type, this.props.model,
+                                 this.props.children);
+    }
+  }
+
+  const view = IFrameView.create('iframe', {
+    opener: new VirtualAttribute((node, current) => {
+      const element = current instanceof String &&
+                      current.unbox &&
+                      current.unbox();
+
+      if (element && node !== element) {
+        for (let {name, value} of node.attributes) {
+          element.setAttribute(name, value);
+        }
+
+        for (let name of node.properties.names) {
+          element[name] = node.name;
+        }
+
+        node.parentNode.replaceChild(element, node);
+      }
+    }),
     isFocused: Focusable.Field.isFocused,
     remote: new BeforeAppendAttribute('remote'),
     mozbrowser: new BeforeAppendAttribute('mozbrowser'),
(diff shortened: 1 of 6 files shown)


src/browser/iframe.js
```diff
@@ -92,7 +137,22 @@
...
+
+  // Gecko API for handling `window.open` has very unfortunate design
+  // (see #566, #565, #564). To workaround it and present more desirable
+  // API that also works with react we employ a hack to generate `opener`
+  // `String` instance (not a primitive value) that holds a reference to
+  // an iframe. It is used to box / unbox actual iframe that is then used
+  // by an IFrame to replace nodes.
+  const Opener = iframe => {
+    // See: https://github.com/mozilla/browser.html/issues/568
+    iframe.isSetVisibleBroken = true;
```
KevinGrandon: You can also wrap setVisible in a try/catch statement as well - whichever you feel better about.

Gozala: > You can also wrap setVisible in a try/catch statement as well - whichever you feel better about.

I'm avoiding them so we will notice if there is another bug like this we don't yet know of.

paulrouget:
We should fix that in gecko.

@KevinGrandon do we know why these frames throw?

src/browser/iframe.js
```diff
@@ -23,14 +68,14 @@
     }),
     isVisible: VirtualAttribute((node, current, past) => {
       if (current != past) {
-        if (node.setVisible) {
+        if (node.setVisible && !node.isSetVisibleBroken) {
```
paulrouget: execute `setVisible` even if `isSetVisibleBroken` is `true`, an catch the exception. When you catch the exception, if `isSetVisibleBroken` just keep going, if `isSetVisibleBroken` is false, throw the exception again.

Something along these lines:

```
if (node.setVisible) {
  try {
    node.setVisible()
  } catch(e) {
    if (!expectIsSetVisibleBroken) {
       throw(e);
    }
  }
}
```

Just in case we magically fix `setVisible` in Gecko and we forget to remove this hack.

Gozala:
Oh I forgot to include added files, will push them in a bit.

src/browser/iframe.js
```diff
@@ -4,9 +4,54 @@
   'use strict';
 
   const Focusable = require('../common/focusable');
-  const {Element, BeforeAppendAttribute, VirtualAttribute, Event, VirtualEvent} = require('../common/element');
+  const {ElementView, BeforeAppendAttribute, VirtualAttribute, Event, VirtualEvent} = require('../common/element');
+  const React = require('react');
+  const {html} = require('reflex');
 
-  const view = Element('iframe', {
+  /*
```
gordonbrander: Should this commented-out code be here?

src/service/force.js
```diff
@@ -0,0 +1,13 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
+
+const {Record, Any} = require('../common/typed');
```
gordonbrander: Should this action live in services?

gordonbrander:
I'm getting an error when I try to click a preview card on this branch:

screen shot 2015-08-12 at 12 22 55

gordonbrander:
Per https://github.com/mozilla/browser.html/issues/475#issuecomment-128927204, should we hold off on landing this until we're sure we can't fix it in Gecko?

src/browser/iframe.js
Gozala: I'll remove it.

src/service/force.js
Gozala: > Should this action live in service directory?

That's a good question, initially it was supposed be a service that deals with this hack https://github.com/mozilla/browser.html/pull/569/files#diff-d0e0be46849a6d8f84059218d61e3fc5R30 but than it ended up inlined for a technical reasons, which is unfortunate but never the less a case. Maybe at this point it indeed would be more appropriate to move it to `common` instead.

Gozala:
> Per #475 (comment), should we hold off on landing this until we're sure we can't fix it in Gecko?

I thin we should land this and remove workarounds once fixes are in gecko, also I am not very optimistic those fixes are gonna happen anytime soon or even sooner that we'll switch to servo.

gordonbrander:
Ok

gordonbrander:
Works.

PR #570, opened 2015-08-08
replace collapse with display:none

### LINKED ISSUE #563: remove all references to visibility:collapse
`visibility:collapse` is not supported by gecko yet (at least for non-table elements), and we use it like `display:none` in some places. Let's get rid of that.

paulrouget:
Fix #563

### DIFF
--- a/src/browser/suggestion-box.js
+++ b/src/browser/suggestion-box.js
@@ -132,7 +132,7 @@
       pointerEvents: 'none'
     },
     collapsed: {
-      visibility: 'collapse'
+      display: 'none'
     },
     suggestions: {
       color: 'rgba(0,0,0,0.7)',


src/browser/suggestion-box.js
```diff
@@ -132,7 +132,7 @@
       pointerEvents: 'none'
     },
     collapsed: {
-      visibility: 'collapse'
+      display: 'none'
```
Gozala: This breaks screenshot feature (See https://github.com/mozilla/browser.html/blob/master/src/browser/web-view.js#L354-L357) #266

Gozala: Maybe `scale(0)` will work, but I have not tried that yet.

paulrouget: Sorry, I don't understand how that's related.

Gozala: Never mind I confused this with web-view collapse.

Issue #538, opened 2015-07-30
ctl tab quickly should switch immediately


gordonbrander:
Currently when hitting `ctl tab` we fade to overview mode. If you command tab very quickly, we should immediately switch pages without the fade.

- [ ] Delay switching to `select-web-view` until `ctl tab` has been held down for 100ms
- [ ] Delay `ctl tab` key repeats until `select-web-view` has been showing for 500ms (we should probably also throttle repeats)
- [ ] Add distinction between WebView `selected` and `highlighted`
- [ ] Select highlighted webview on `SynthesisUI.Select` completion. Then fade to `show-web-view`.
- [ ] If `ctl tab` keys are released prior to 100ms delay, don't show `select-web-view`. Select next web view immediately.

^ The same obviously applies to `ctl shift tab`.

paulrouget:
You mean `Ctrl-Tab`?

gordonbrander:
Yes, sorry.

gordonbrander:
We need to do something like:

                             *-> delay interrupted by keyup -> switch to prev tab immediately
                            /
    ctl tab -> delay 30ms -*
                            \
                             *-> delay complete -> fade to select-web-view mode

PR #572, opened 2015-08-11
Implement immediate tab switcher #538

Gozala:


paulrouget:
This doesn't work as explained in #538:

> Delay switching to select-web-view until ctl tab has been held down for 100ms

Gozala:
> This doesn't work as explained in #538:

Me and @gordonbrander have being chatting about this in slack, this pull was in the intermediate state at the time of your comment. Sorry should have pointed out in the pull.

Gozala:
@gordonbrander I think it is complete now, please make a thorough review to make sure I did not regress anything else. @paulrouget feel free to also do the review more eyes can only help preventing regressions.

src/browser/synthesis-ui.js
```diff
@@ -127,73 +123,101 @@
...
   exports.update = update;
+
+
+  const service = address => {
+    let id = -1;
+    const showPreview = address.pass(ShowPreview)
+
+    return action => {
+      if (action instanceof WebView.Preview) {
+        id = setTimeout(showPreview, 70);
```
gordonbrander: I suggest a slightly longer timeout. It's still possible to see a flicker of the switcher if I don't release tab fast enough.

src/browser/synthesis-ui.js
```diff
@@ -46,18 +56,12 @@
   const clearInput = edit('input', Editable.clear);
 
 
+  const selectViewByIndex = (state, index) =>
```
gordonbrander: should this be `previewByIndex`?

src/browser/web-preview.js
```diff
@@ -20,6 +25,7 @@
     return null;
   }
 
+
```
gordonbrander: 😕

gordonbrander:
...other than my small comments, looks good.

src/browser/synthesis-ui.js
Gozala: Nope, notice that `WebView.activate(...)` it makes `x.selected = x.previewed` and we set `x.previewed` so that two don't get out of sync when you for instance click a card to select it.

src/browser/synthesis-ui.js
Gozala: > I suggest a slightly longer timeout. It's still possible to see a flicker of the switcher if I don't release tab fast enough.

Can you play around with numbers and suggest one that you find to be the best, in my comparison to OSX app switcher this seemed the closest  but a lot of variables can be in a play here like the power of machine personal perception etc... I'm happy to set this to whatever you find to be the best.

src/browser/synthesis-ui.js
gordonbrander: Sounds good.

gordonbrander:
Let's land this and file a follow-up bug to tune the timeout length.

PR #573, opened 2015-08-11
Make settings editable.

Gozala:


Gozala:
@paulrouget can you please take a look when you'll get a chance ? Not an important patch, just a convenience thing.

src/about/settings/index.js
```diff
@@ -28,7 +55,99 @@
...
+    },[
+      html.span({
+        key: 'sizer',
+        style: {
+          left: 0,
+          color: 'red',
+          whiteSpace: 'pre',
+          opacity: 0,
+        }
+      }, props.value),
```
paulrouget: Why do you need that?

src/about/settings/index.js
```diff
@@ -28,7 +55,99 @@
...
+      input({
+        style: {
+          padding: 0,
+          margin: 0,
+          background: 'transparent',
+          border: 'none',
+          font: 'inherit',
+          margin: 0,
+          padding: 0,
```
paulrouget: Already defined.

src/about/settings/index.js
```diff
@@ -28,7 +55,99 @@
...
+          whiteSpace: 'pre',
+          opacity: 0,
+        }
+      }, props.value),
+      input({
+        style: {
+          padding: 0,
+          margin: 0,
+          background: 'transparent',
+          border: 'none',
```
paulrouget: Keep the original look. That makes it difficult to use.

src/about/settings/index.js
```diff
@@ -28,7 +55,99 @@
...
+    html.input({
+      type: 'checkbox',
+      checked: value,
+      onChange: event => address.receive(Settings.Update({
+        name, value: event.target.checked
+      }))});
+
+  const viewString = (name, value, address) =>
+    html.code(null, [
+      '"',
```
paulrouget: If you want to make it clean it's a string, do not use `"`, keep the native `<input>` style.

src/about/settings/index.js
Gozala: > Why do you need that?

I want input field to resize with the size of input, to achieve that input width is 100% and positioned absolute, while there is also a hidden span that has same style as input but isn't absolute there for causing it's parent to resize.

src/about/settings/index.js
Gozala: Well you still may change the type if you say edit `null` field to `"null"` it will become `"null"` instead. But anyway maybe you're right and I did over engineered this.

src/about/settings/index.js
```diff
@@ -28,7 +65,125 @@
 
   // View
 
-  const viewSetting = (name, value) => html.div({
+  const fieldStyle = StyleSheet.create({
+    // input field is wrapped in a box so that input field itself
+    // will resize to match it's container & subsequently typed text.
+    // input field is positioned absolute so it's size won't affect
+    // container & it's styles are blanked out as container is supposed
```
paulrouget: s/it's/its. x2

src/about/settings/index.js
paulrouget: Why not use contentEditable?

paulrouget:
Can we do that in 2 steps, first make about:settings editable with a basic style (no dark theme, not fancy inputs), and then see with @gordonbrander how we could make this page better?

----

I've moved all the settings we might want to edit in the devtools HUD. Which is kinda of ugly.

If we could merge the devtools code in this about:settings page, that'd be great. But that would also mean that we need to add some special rendering for some properties (like the one to enable the devtools).

Gozala:
> Can we do that in 2 steps, first make about:settings editable with a basic style (no dark theme, not fancy inputs), and then see with @gordonbrander how we could make this page better?

I started with that but I plain inputs for editing made editing really unpleasant so I tried to improve it by making more editor like. Anyway I can switch back to the plain inputs fields and we can improve this in the next iteration.

> I've moved all the settings we might want to edit in the devtools HUD. Which is kinda of ugly.

> If we could merge the devtools code in this about:settings page, that'd be great. But that would also mean that we need to add some special rendering for some properties (like the one to enable the devtools).

I am not totally sure what the suggestion here ? Edits in the devtools HUD and `about:settings` now are synced right away so you should be able to use just one of them. Unfortunately settings don't come with natural language descriptions like you see in devtools HUD and I'm not entirely sure how such descriptions could be added without extending settings API. One though I had was to employ `describe:layers.dump` style setting names for storing descriptions which we could later use in the UI to display better descriptions for settings but that seems like a hack and I'm not totally sure that doing that is even worth it. For devtools hud it may make sense to just create `about:debugging` page or something to display those.

paulrouget:
>  Anyway I can switch back to the plain inputs fields and we can improve this in the next iteration.

Yes please. r+ for the non-style code.

Let's try to get the style right in another PR.

PR #588, opened 2015-08-13
Fix #574 regressions affecting history-db

### LINKED ISSUE #574: Bunch of promise failures come from history-worker
I see bunch of errors being logged about unhandled rejected promises in the console, we should look into them and get rid of those errors.

Gozala:
Fix for #574


gordonbrander:
Works.

PR #596, opened 2015-08-14
Implement top hit suggestion per spec (Fix #559)

### LINKED ISSUE #559: Implement "top hit" for Awesomebar
Per the Synthesis UI design.

- In addition to history and search suggestions, search for root domains in history that match search term (e.g. facebook.com).
- If we have a match, style it as a "top hit".

This is to help people get back to their favorite websites.

Gozala:


gordonbrander:
Behavior looks good, but there are some problems:

- The spacing around search list items is no longer correct. Icons and text no longer line up with search box icon.
- Favicons are inconsistently sized. We should do as we do for the card favicons by setting them as `background-size: cover` on a fixed-size `<div>`.
- When we have no favicon, there is a blank space. We should display text flush to the left when there is no favicon.

@Gozala if you like, I can update your branch and submit an updated PR

Gozala:
@gordonbrander how about we land this and you make a followup with style changes ?

gordonbrander:
Styling is incorrect, but I'll land and follow up with style fixes.

PR #597, opened 2015-08-14
Fix regression in cmd-click introduced by #569

### LINKED ISSUE #592: cmd click is broken
It no longer works. I wonder if this is fallout from https://github.com/mozilla/browser.html/pull/572?

Gozala:
Fix #592


gordonbrander:
@Gozala this does fix it, but as I was testing it I found that it's weirdly not working specifically on http://www.theverge.com/. There doesn't seem to be anything special about The Verge's links :/

gordonbrander:
It works. The problem with the Verge is platform-level bug. We're going to open a separate bug for it.

PR #598, opened 2015-08-14
add github to curated sites list.

Gozala:
Let's add a colors for github to the curated list otherwise their theme color is just annoyingly odd and makes us feel stupid.


gordonbrander:
Ok

PR #615, opened 2015-08-19
Reconcile edit/create states

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/556.

Missing from this commit:

- Dynamically re-positioning tabs based on whether ghost tab is present. I think we need https://github.com/mozilla/browser.html/issues/577 for that.

Description:

create-web-view
- Cards are listed left-to-right in create-web-view (maybe it should be right-to-left, but this creates a weirdness with ctl tab. Maybe that's ok.)
- Focusing input will take you to the head of the list. Ghost card will appear.
- Unfocusing will make ghost card disappear and return viewport to where you were.
- Click plus takes you here. Input is focused.
- pinch/zoom from `show-web-view` takes you here, centered on current card. Input is not focused. Other than centering, there is no selected state.
- If the location bar has a value, we switch to search-view. You're conceptually editing the ghost tab. Hitting 'enter will fill ghost tab'.
- Closing a page will take you to create-web-view, centered on the next webview in the list (if any). The location bar is not focused unless there are no webviews.
- cmd clicking links will open links next to the page you're on, rather than at the head of the list (it's like rewriting history). They're opened in the background.

search-results
- edit-web-view is replaced by search-results view
- Search/edit is a full-window experience
- Clicking the 'x' button in the search box will return you to view-web-page of the selected web view, or to `create-web-view` if there are no web views.
- Clicking the location bar will take you to search mode. Unfocusing will take you back to page. Hitting enter will change tab location.

select-web-view
- This view is visually the same as `create-web-view`, except that input is never focused and ghost card is never shown.
- Selected style is identical to hover style

Motivation:

- Creates an obvious distinction between create view and edit view. Resolves the state edge cases currently present.
- In future, this will allow us to tie page-specific search-in-page into search box, since clicking on minimized location bar will transition straight to search view, where we can have suggestions about scope.

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -25,6 +25,14 @@
   visibility: visible;
 }
 
+.card {
+  transition: transform 80ms linear;
+}
+
+.card:hover {
+  transform: scale(1.05);
+}
+
 * {
   -moz-appearance: none;
   background-color: transparent;
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -179,7 +179,7 @@
       render('LocationBar', LocationBar.view,
         state.mode, loader, security, page, input, suggestions, address),
       render('Preview', Preview.view,
-        state.mode, webViews.loader, webViews.page, webViews.previewed, theme, address),
+        state.mode, webViews.loader, webViews.page, input.isFocused, webViews.previewed, theme, address),
       render('Suggestions', Suggestions.view,
         state.mode, suggestions, input, address),
       html.div({
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -179,7 +179,7 @@
       render('LocationBar', LocationBar.view,
         state.mode, loader, security, page, input, suggestions, address),
       render('Preview', Preview.view,
-        state.mode, webViews.loader, webViews.page, input.isFocused, webViews.previewed, theme, address),
+        state.mode, webViews.loader, webViews.page, input.isFocused, webViews.previewed, address),
       render('Suggestions', Suggestions.view,
         state.mode, suggestions, input, address),
       html.div({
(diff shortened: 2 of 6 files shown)


Gozala:
I have not went through code yet, but here is my initial impressions of using this branch:

- Focusing address bar causes odd flickering, you can see a dashboard for a split second and then then you go to white screen (presumably search-mode).
- Switching tabs (via `ctrl tab`) causes odd redraw just before tab is activated. Possible this happens in current master as well but probably because of style changes is noticeable.
- Zooming out of tab seems to take me to a confusing place. It's like tab switcher but no tab is selected. If I focus input bar "ghost tab" appears. If I'm taken to the white screen (presumable search mode) then if I load page it loads and enters page. So many screens jumps I feel little nausea. Now if I zoom out again I'm taken to the same place as before but this time input is focused and ghost tab appears right away.
- Now tab switch mode seems different than zoom out mode since you do not have any card scaled. If they are different maybe removing search bar from tab switch mode is better because you can't really use it and just seems more confusing. I think it was less so because in the past at least visually those two modes were identical.
- In the current implementation I frequently use `cmd l` to get an overview of the open / active pages. Now I get white screen instead that I don't seem to enjoy that much. As far as I can tell only way to get an overview now (without switching) is via zoom-out gesture. But then I no longer see which page is active and I personally prefer to use keyboard.
- `ctrl tab` in `zoom-out` screen causes an odd jump presumably because ghost tab is not present there, it would be worth it to still keep it there but hide it or at least add some margin to compensate it's absence and to avoid jumping cards.
- I kind of wish all the dashboard like screens were the same and ghost tab was also always there so you could select it to enter an address like it used to be in previous design. To be honest with this changes I feel more confused and lost then without them.

Issue #560, opened 2015-08-06
Drag / 2-finger flick tabs to close


gordonbrander:
For interaction let's try:

- Swipe up or down by a fixed `threshold` to close card (a threshold of 200 is probably a good place to start, but we should make this tweakable)
- If user releases inside `threshold`, card moves back into place.
  - Good: we transition back using a fixed-duration CSS animation
  - Better: we base transition duration on avg velocity of swipe upon release, but will add accel to velocity until we reach `minimum` if velocity is too slow.
- When user passes `threshold`, but has not yet released, card transitions to semi-transparent using fixed CSS transition. If user moves card back within threshold, card transitions to opaque.
- When user releases finger(s) from trackpad past threshold, we trigger the close animation:
  - Card moves in direction of swipe, at swipe velocity (we add accel until velocity is at `minimum`, if velocity is too slow)
  - At the same time, card fades out. Fade progress should be tied to distance. I think we want card to be "gone" visually at about `100px` past `threshold`. We may want to make this tweakable too.
  - When card is gone, cards to the left and right move in to fill space, using a fixed-duration CSS transition.

Gozala:
Depends on #571

PR #617, opened 2015-08-20
Work in progress implementation of flick to close gestures

Gozala:
Fix for #560


Gozala:
Ok current implementation is little different from the spec:

1. `MozSwipeGesture` events don't get fired (see https://github.com/mozilla/browser.html/issues/571#issuecomment-133168532) which makes it impossible for us to know when user realized touch. To workaround that `releaseDistance` constant was introduced. If card travels that much pixels past `threshold` it will behave if user have released a touch.
2. Turns out that fading and closing card at the same `threshold` does not feel right, so separate `fadeDistance` and `dropDistance` constants have being introduced. Card starts fading as soon as it passes `threshold` and it reaches complete transparency `fadeDistance` pixels passed `threshold`. Opacity is proportional to a distance from `threshold`. Card can move at most `dropDistance` pixels from `threshold` (by then it's completely transparent since `fadeDistance` is less) once it passed that distance card placeholder starts to  shrink via css transition providing "fill space" effect. After `10ms` since shrinking card is marked shrinked which triggers close event.
3. CSS transition for opacity change is commented out, that's because transition timing and opacity changes did not quite synced well. It could be that choosing carefully transition duration and `fadeDistance` would have worked, but I was unable to find the right numbers & even if I did it probably would feel different on other machine so I chose to not use transition.
4. Average swipe velocity with fixed minimum was not used, because depending on how you swipe you could get odd behavior (card slowing down or speeding up too much when you change speed during swipes). Instead overall acceleration is calculated through out the swipe and then is used to for calculating velocity (with constant unrealistic upper bound) that is used to calculate position. This seems to feel far more natural.

I still would like to take some time to cleanup the code as I expect there maybe some redundant calculations. I am also not very happy with a complexity of code, but I hope to find a way to simplify during cleanup or post release.

gordonbrander:
- [ ] The swipe-up/swipe-down animation is janky, with jerky movement. This may be GC-related pauses, or it may be that we need to average the velocity of the touchmoves over `dt`.
- [ ] The "closing the gap" animation where the cards move together to fill the space that was left by the closed card is very janky. Sometimes it skips the animation altogether.

src/browser/web-card.js
```diff
@@ -0,0 +1,108 @@
...
+  return state.timeStamp === null ? state :
+    state.isClosing ?
+      state.merge({
+        elapsedTime,
+        beginShrink: state.beginShrink > 0 ? state.beginShrink :
+                     state.y === maxY || state.y === -1 * maxY ? action.timeStamp :
+                     state.beginShrink,
+        endShrink: state.endShrink > 0 ? state.endShrink :
+                   state.beginShrink - action.timeStamp > 10 ? action.timeStamp :
+                   state.beginShrink,
```
Gozala: This supposed to be a `state.endShrink` instead.

src/browser/web-card.js
```diff
@@ -0,0 +1,108 @@
...
+  const maxY = thresholdY + dropDistance;
+  return state.timeStamp === null ? state :
+    state.isClosing ?
+      state.merge({
+        elapsedTime,
+        beginShrink: state.beginShrink > 0 ? state.beginShrink :
+                     state.y === maxY || state.y === -1 * maxY ? action.timeStamp :
+                     state.beginShrink,
+        endShrink: state.endShrink > 0 ? state.endShrink :
+                   state.beginShrink - action.timeStamp > 10 ? action.timeStamp :
```
Gozala: Should not hard code 10 instead constant should be defined for the duration.

src/browser/web-card.js
```diff
@@ -0,0 +1,110 @@
...
+        velocity: action.y / elapsedTime,
+        acceleration: (action.y / elapsedTime) / elapsedTime
+      }) :
+    state;
+};
+
+const release = (state, action) =>
+  state.y > thresholdY ? state.set('isClosing', true) : state;
+
+export const update = (state, action) => {
```
Gozala: @gordonbrander can you please add following code here and post output somewhere so we can get some insight of what's going on:

```js
console.log('------>')
console.log(state.toJSON())
console.log(action.toJSON())
console.log('------<')
```

src/browser/browser.js
```diff
@@ -4,7 +4,7 @@
   'use strict';
 
   const {html, node, render, cache} = require('reflex');
-  const {Record, Any, Union, Maybe} = require('../common/typed');
+  const {Record, Maybe} = require('typed-immutable');
```
gordonbrander: Did a dependency update? Does this also need to change in all the other modules?

Gozala: > Did a dependency update? Does this also need to change in all the other modules?

I should probably make a separate pull for those changes.

src/browser/web-card.js
```diff
@@ -0,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+'use strict';
+
+import {Record, Union, List, Maybe, Any} from 'typed-immutable';
```
gordonbrander: :+1: for ES Modules. Is this going to cause any problems with other modules that are AMD?

src/browser/browser.js
Gozala: I have submitted bug for this #625 and will factor out those changes into separate pull.

src/browser/web-card.js
Gozala: > for ES Modules. Is this going to cause any problems with other modules that are AMD?

babel compiles everything into UMD so after compile step it's all fine. If you would to try and load this with requirejs it will fail because this syntax is not even implemented by a browsers. Although all other modules will also fail with requirejs because we no longer wrap them into `define(...)` as that was confusing bundler.

gordonbrander: Ok, nvm.

src/browser/web-preview.js
```diff
@@ -66,13 +69,16 @@
...
       margin: '0 10px',
       overflow: 'hidden',
       position: 'relative',
-      'scrollSnapCoordinate': '50% 50%',
-      width: '240px'
+      width: '240px',
+      top: '50%',
+      marginTop: '-50%',
+      // Seems to make things worse
+      // transition: 'ease-in opacity 0.1s'
```
gordonbrander: We should get rid of this commented-out code.

src/browser/web-preview.js
```diff
@@ -66,13 +69,16 @@
...
+      transition: 'ease box-shadow 0.15s'
```
gordonbrander: Animations are usually `<property> <time> <easing function>` but CSS is liberal in what it accepts, so this works.

src/browser/web-preview.js
```diff
@@ -246,20 +314,29 @@
       // Margin doesn't play well with scroll -- the right-hand edge will get
       // cut off, so we turn on the traditional CSS box model and use padding.
       boxSizing: 'content-box',
-      width: '100vw',
+      height: '100%',
       // Fixed height to contain floats.
-      height: '300px',
-      padding: 'calc(50vh - 150px) 200px 0 200px',
+      //padding: 'calc(50vh - 150px) 200px 0 200px',
```
gordonbrander: We should get rid of this commented-out code.

src/browser/web-security.js
```diff
@@ -23,6 +23,9 @@
   }, 'WebView.Security.Change');
   exports.SecurityChanged = SecurityChanged;
 
+  const Action = SecurityChanged;
```
gordonbrander: Is this change related?

PR #620, opened 2015-08-20
Allow location edit for different domains

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/576.

@paulrouget r?

### DIFF
--- a/src/browser/synthesis-ui.js
+++ b/src/browser/synthesis-ui.js
@@ -7,7 +7,7 @@
   const {compose} = require('../lang/functional');
   const WebView = require('./web-view');
   const Preview = require('./web-preview');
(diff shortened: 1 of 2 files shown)


src/browser/synthesis-ui.js
```diff
@@ -114,7 +115,7 @@
     const webViews = state.mode === 'edit-web-view' ?
       WebView.loadByIndex(state.webViews,
                           state.webViews.selected,
-                          {uri}) :
+                          LocationChanged({uri, timeStamp: Date.now()})) :
```
Gozala: You should not use `LocationChanged` (that is action that occurs when user navigates away by clicking url) but rather `Loader.Load({uri})`.

PR #636, opened 2015-08-25
Bump a version as card animations was incompatible.

Gozala:


Gozala:
@paulrouget @gordonbrander After this change we should bump last version digit on every state structure change. On a release we should increment middle digit and reset last digit to `0`.

PR #640, opened 2015-08-26
make browser.html more servo compliant

paulrouget:


src/common/editable.js
```diff
@@ -73,10 +73,12 @@
...
-        const {start, end, direction} = current;
-        node.setSelectionRange(start === Infinity ? node.value.length : start,
-                               end === Infinity ? node.value.length : end,
-                               direction);
+        if (node.setSelectionRange) { // FIXME: remove once Servo supports setSelectionRange
+          const {start, end, direction} = current;
+          node.setSelectionRange(start === Infinity ? node.value.length : start,
+                                 end === Infinity ? node.value.length : end,
+                                 direction);
+        }
```
Gozala: Does servo support `node.selectionStart = n` if so we may use that if `node.setSeletionRange` is not defined.

paulrouget:
> Does servo support node.selectionStart

No.

PR #641, opened 2015-08-27
Fix the race condition with incorrect loads.

### LINKED ISSUE #618: Loading first site fails
Steps to reproduce:
- get an empty profile.
- launch graphene --start-manifest=...
- the focus is on "Search or enter address..."
- enter a website address, eg mozilla.org, and hit enter.

Expected:
We load the page

Observed:
We load the "File not found. :(" page.

Gozala:
Supposed to fix #618

### DIFF
--- a/src/browser/iframe.js
+++ b/src/browser/iframe.js
@@ -4,54 +4,50 @@
   'use strict';
 
   const Focusable = require('../common/focusable');
-  const {ElementView, BeforeAppendAttribute, VirtualAttribute, Event, VirtualEvent} = require('../common/element');
+  const {Element, BeforeAppendAttribute, VirtualAttribute,
+        BubbledEvent, VirtualEvent} = require('../common/element');
   const React = require('react');
   const {html} = require('reflex');
 
-  class IFrameView extends ElementView {
-    componentDidMount() {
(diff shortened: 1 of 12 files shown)


Gozala:
Waiting on @fabricedesre to confirm this resolves an issue.

src/browser/iframe.js
```diff
@@ -4,54 +4,50 @@
...
     opener: new VirtualAttribute((node, current) => {
-      const element = current instanceof String &&
-                      current.unbox &&
-                      current.unbox();
-
-      if (element && node !== element) {
-        for (let {name, value} of node.attributes) {
-          element.setAttribute(name, value);
-        }
+      const element = current instanceof String && current.unbox != null ?
```
gordonbrander: Why check against null here? We need a function.

gordonbrander:
It looks good to me. Let's resolve conflicts and merge.

src/browser/iframe.js
Gozala: I should not have, but also played with http://flowtype.org that's why there all this type annotations. Short story is flow can detect those things and in this case type is `() => any|void` so it will not type check if you don't make sure it's not non `nil` :)

PR #646, opened 2015-08-29
Implement page slide based navigation.

Gozala:
@gordonbrander this is an attempt to implement page slide based go back / forward, but unfortunately it's unusable as screenshots arrive just too late, usually when the sliding page is about slide out of viewport. And when it does it render image it causes junk.

P.S.: I have not actually hooked up navigation events because idea was to trigger goBack after we have a screenshot but this just unusable to be worth finishing.

Related to #646

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "browser.html",
   "description": "browser.html",
-  "version": "0.0.8",
+  "version": "0.0.9",
   "homepage": "https://github.com/mozilla/browser.html",
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


paulrouget:
I suggest that we put that on hold until we have a better understanding of how the browser API should evolve.

Taking hidpi screenshots of a page (child process) and sending it to the main process, without really knowing when to take the screenshot (which even?) will always be inefficient.

At the platform level, we can certainly make this much better. Let's just make sure we know what effect we want to achieve, how that will work with the Browser API, and fix it platform side.

Issue #502, opened 2015-07-16
Implement back/fwd 2 finger swipe gestures


gordonbrander:
Part of sythesis UI that has not yet been implemented.

gordonbrander:
We can use this https://developer.mozilla.org/en-US/docs/Web/Events/MozSwipeGesture

paulrouget:
I'm not sure why, but swipe gestures don't work anymore. I'll investigate.

gordonbrander:
We should slide one page over the other while you swipe.

paulrouget:
Like Safari?
And it goes through history, not through tabs, right?

gordonbrander:
> Like Safari?
> And it goes through history, not through tabs, right?
> 
Right.

gordonbrander:
Sliding the new page over the old page is definitely the correct design decision, but it will probably need platform support. Options:

- We can defer for the MVP (https://github.com/mozilla/browser.html/milestones/Developer%20preview%20(Gecko%20Graphene)) and revisit in Servo.
- Slide over blank page, then render when slide is complete.

gordonbrander:
Blocked by https://github.com/mozilla/browser.html/issues/571

PR #650, opened 2015-09-01
Implement tilt swipe animation for navigating back / forward 

Gozala:
Fix for #502 that is alternative design for #646


gordonbrander:
R+. It's pretty good, but not perfect. Let's follow up with some tweaks to the animation.

Issue #623, opened 2015-08-21
When search results re-render, they blow away user selection


gordonbrander:
We should make sure not to do that. I've more than once tabbed down to a top hit, only to have it disappear just before I hit enter, and some other random site opens in its place. cc @Gozala

PR #656, opened 2015-09-03
Retain selected suggestion across suggestion updates.

Gozala:
Fix for #623

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "browser.html",
   "description": "browser.html",
-  "version": "0.0.9",
+  "version": "0.0.10",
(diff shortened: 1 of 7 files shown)


Gozala:
- [ ] Need to increment a version as db model has changed.

gordonbrander:
Interaction bug:

- Arrow down to top hit
- Select all (cmd a) and then start typing

What should happen: blow away old text and insert new text.
What happens: the url of the top hit stays in the input even as you type. Cursor jumps around.

gordonbrander:
Apparently it works this way in master. @Gozala to file a bug for it.

gordonbrander:
R+. @Gozala can you bump the db model version and then merge?

Issue #644, opened 2015-08-28
App should start only after build is finished


Gozala:
This requires some digging into [gulpsj](http://gulpjs.com) which I'm not at all familiar with. As far as I can tell gulp runs dependent tasks concurrently so sometimes browser app will lunch before dist is created and there for it fails to load. It should first finish build and only then lunch an app.

gordonbrander:
@Gozala I believe this is the case, no?

gordonbrander:
This is done (see above).

Issue #663, opened 2015-09-09
transaction monitor


Gozala:
Redux devtools is neat tool that logs every action and state modification that you can manually interact with:
https://github.com/gaearon/redux-devtools

You can see the demo of this tool in action https://youtu.be/xsSnOQynTHs?t=17m22s

In this issue I'll try to integrate redux-devtool or create a similar tool (please note that our setup is little more complex than demoed and we don't actually use redux either so it could be that creating a tool inspired by it would be easier than integrating it).

Please also note that one of the key features is changing a code and then replaying for which we would need to get #662 going first.

PR #665, opened 2015-09-09
Implementation of initial live code swapping

Gozala:
[react-hot-loader](https://gaearon.github.io/react-hot-loader/) is a popular tool in react community that enables developers to apply code edits on the running app without reload. While this is useful on it's own it's a foundation for working with transaction-monitor (see #663 for details) that allows you to record session and then replay each step, with live code swapping you can fix a bug and reply steps to make sure it's fixed.

Now current setup uses [livereactload](https://github.com/milankinen/livereactload) which is clone of [react-hot-loader](https://gaearon.github.io/react-hot-loader/) for [browserify](http://browserify.org) instead of [webpack](http://webpack.github.io) which I chose to use because:

1. We already used browserify for doing builds so it's was easier to integrate.
2. Unfortunately react-hot-loader did not worked for us out the box as some of the assumption it makes were not valid in our setup.

Ultimately I think switch to react-hot-loader and webpack would be beneficial as unlike livereactload react-hot-loader can update individual modules rather than whole bundle so it's going to be a lot faster.

## Known limitations / issues

1. iframes are reloaded when code is swapped (Caused by react issue facebook/react#4826)
2. changes to about:settings do not cause code swap. This is actually intentional since just reloading `about:settings` page seems good enough.

## Usage

Currently to use this you need to run `npm run live` or if you just interested in server and not a client you can run `npm run live-server`. Please note that we need to have eval to enable hot swapping there for `.profile/user.js` file was added to enable eval for development. If you run client separately make sure to use profile that will have a similar setup.


Gozala:
This also fixes #644

gordonbrander:
If we need eval, how will this work in Graphene?

paulrouget:
Because it's only for development, we can allow eval.

Gozala:
> If we need eval, how will this work in Graphene?

As @paulrouget said it's only enabled for development so code can be live reloaded. Live reloading is not part of the deployment build so no eval is required there.

gordonbrander:
Looks like we don't watch for changes to the CSS directory. We probably should, right?

gordonbrander:
Other than that, looks good.

Gozala:
> Looks like we don't watch for changes to the CSS directory. We probably should, right?

We should probably just kill css directory. Either way live reload knows how to hot swap js code it can't really do anything with css. We could teach it how to swap css as well but I don't think it's worth it given that it's almost not used.

gordonbrander:
Ok, wfm.

PR #666, opened 2015-09-11
Livereload iframes

Gozala:
Followup fix for #665 that avoids reloading iframes on code hot-swap. It also takes care of hot-swapping services (without this fix services that registered listeners kept alive the whole copy of the app, for example swipe gesture would took completely took over).


src/browser/iframe.js
```diff
@@ -6,7 +6,7 @@
   const Focusable = require('../common/focusable');
   const {Element, BeforeAppendAttribute, VirtualAttribute,
         BubbledEvent, VirtualEvent} = require('../common/element');
-  const React = require('react');
+  const React = require('../common/react');
```
gordonbrander: Did we fork react?

src/browser/index.js
```diff
@@ -46,6 +46,7 @@
       settings(action);
       scraper(action);
       synthesis(action);
+      gesture(action);
```
gordonbrander: Should this be part of this commit?

gordonbrander: Eh, nvm. Good enough.

PR #681, opened 2015-10-22
virtual-dom integration work

Gozala:
# Do not land yet!

### High level overview of the changes

- [livereactload][] is by [browserify-hmr][] for couple of reasons:
  1. [livereactload][] was very react specific doing bunch of code transforms to handle hot-swap react components and failed to do it properly for non-react components.
  2. [livereactload][] is a port of [react-hot-loader][] that [got deprecated](https://github.com/gaearon/react-hot-loader#deprecation-notice) in favor [webpack-hmr][] + [react transform][], which is more or less what we do here although we use [browserify-hmr][] instead (since we're using browserify) and [hotify][] transform instead. Latem makes more sense for our architecture since we can reload whole app without monkey patching react components, also in this case we don't even use react or it's components.
  3. Unlike  [livereactload][] which rebuild bundle on every change [hmr][webpack-hmr] rebundles only changes and sends those, which means changes to worker code just reloads a worker. Also given smaller chunks overall feedback loop is significantly faster.

- legacy `about/dashboard` code was removed, in fact it was not used and is not specific to this pull.
- In new reflex `address` is no longer an object but rather a function, there for:
  1.  Method `address.receive(action)` is replaced by more simple `address(action)`.
   2. Method `address.forward(annotate)` is gone, instead reflex provides `forward(address, annotate)` function that does exactly the same thing. Although since addresses are now functions `forward` also replaced  `address.pass(annotate)` with `forward(address, annotate)`.
   3. Method `address.pass(annotate, arg1, arg2)` is gone. It was rarely used anyway, if you need it just use `forward(address, partial(annotate, arg1, arg2))` or `forward(address, action => annotate(arg1, arg2, action))`
   4. Method `address.send(action)` is also gone. Again it was rarely used and did not really offered much. In most cases you better off defer action instantiation until event happens, still if you have a use case for `.send` you achieve same with `partial(address, action)`.

  > Note: Above mentioned `partial` is defined in `lang/functional` module.
  
 
- Function previously called `render` was renamed to `thunk` to better match it's intent. `thunk(cacheKey, view, ...args)` creates lazy computation for `view(...args)` that is only re-computed if `...args` have changed from previous call, otherwise result of previous computation is re-used. `cacheKey` is used to identify computation as same `view` function can be used over the collection of items & there for cause unnecessary cache invalidations.
- In new reflex when you create a node, last optional argument is it's children. Unlike before children must be an array (previously non-array was treated as array of that element).
- In new reflex when you create a node, last optional argument must be an array. Iterators used to work before but were unintentionally broken gozala/reflex#37 because virtual-dom did not supported them. I will fix that but for now we just spread lists into arrays.

- Even though reflex makes render drivers pluggable we still have some exotic use cases that have not being abstracted. There for for now we have `driver` directory that provides driver specific implementations for exotic things. All of those in case of virtual-dom can be / are implemented using [hooks][] & in a future I'd like to add [hooks][] support to react driver gozala/reflex-react-driver#2 In the meantime:

  1. Non stantard attributes used to be a nightmare to deal with in react. With virtual-dom it's pretty easy: `html.iframe({ id: id, src: location, attributes: { remote: true, mozbrowser: true })` This does not yet works in react driver but intend to fix that gozala/reflex-react-driver#1

  2. Non standard events used to be extremely difficult to deal with react. With virtual-dom it's no brainer using [hooks][] \o/. Now driver exports `on(address, decoder, options, getTarget)` function where all args are optional except an `address`. This can be used to set any non standard listener: `{onMozBrowserLocationChange: on(address, eventToAction)}` creates an event listener for `mozbrowserlocationchange` event where event is decoded to an action with `eventToAction` and then send to an `address`. If `decoder` is omitted `undefined` is send to address this comes handy for with forward: `on(forward(address, Close))`. In some cases event names are not all lower case though for such events it's possible to use same `on` API but property name should not start with `on` in which case property name will be used as an even type: `({MozSwipeGesture: on(address, readGesture)})` registers event listener for `MozSwipeGesture` event.

  3. An odd `animation(node, onNextFrame, isAnimating)` is replaced with another event like hook: `({onFrame: onAnimationFrame(address, decoder)})`. Similar to `on` although property name is irrelevant. address will be passed action created as `decoder(timeStamp)` as long such property is set. **Note:** this is temporary though animations will be dealt with effects instead, but that requires more refactoring.

  4. meta properties like `isVisible`, `isFocused` etc... were treated specially in React. This handled by [hooks][] in virtual-dom driver case. now we have `({isFocused: focus(state.isFocused), isVisible: visibility(state.isVisible), ..})`

- It turns out react and virtual-dom deal with style slightly differently. For example `{height: 50}` is interpreted by react as `height: 50px` while in case of virtual-dom it is interpreted as is meaning style rule is invalid. There for all styles should be written with units: `{height: '50px'}` which works in both rendering drivers and is kind how you do it in css anyhow.

- Sending a `null` to an address used to be a noop. Now `null` is actually send (That's because addresses are simple functions now) which fails some of the code since `null` actions are not expected. In those cases I used `{}` as a workaround but should probably have special action type for that.

### Todo:

- [x] Loading a new tab causes frame drop and runs for more than `150ms` that means something is either wrong with timing or something is going wrong there. Unfortunately profiler seems broken in devtools but I'll investigate what is going on there.
- [ ] Rewrite animations with effects
- [ ] Replace IO services with effects
- [x] Swiping up or down seems to close tab, likely swipe events somehow reach cards beneath the iframe. Need to fix this.
- [x] Escape in location bar does not seem to return focus to the iframe.
- [ ] about:settings must be updated
- [x] Fix devtools-hud
- [x] Fix tab disappearing on tab switching

  P.S.: The way we used to sample was less precise than now, it could be that creating new iframes used to be always slow we just did not captured that in samples. 

[livereactload]:https://github.com/milankinen/livereactload
[browserify-hmr]:https://github.com/AgentME/browserify-hmr
[react-hot-loader]:https://gaearon.github.io/react-hot-loader/
[webpack-hmr]:https://webpack.github.io/docs/hot-module-replacement.html
[react-transform]:https://github.com/gaearon/react-transform-boilerplate
[hotify]:https://github.com/gozala/browserify-hotify
[hooks]:https://github.com/Matt-Esch/virtual-dom/blob/master/docs/hooks.md


gulpfile.babel.js
```diff
@@ -166,7 +183,7 @@ gulp.task('watch', [
 gulp.task('develop', sequencial('watch', 'server', 'application'));
 gulp.task('build-server', sequencial('watch', 'server'));
 
-gulp.task('live', ['livereloader', 'develop']);
-gulp.task('live-server', ['livereloader', 'build-server']);
+gulp.task('live', ['hotreload', 'develop']);
+gulp.task('live-server', ['hotreload', 'build-server']);
```
Gozala: This changes are mainly to simplify live reload handling & to make it significantly faster.

src/browser/index.js
```diff
@@ -19,43 +19,67 @@
...
-      }
-    }
+
+
+  const application = start({
+    initial: isReload ?
+                Browser.Model(window.application.model.value) :
+                Session.restore(version, Browser.Model),
+    update: Browser.update,
+    view: (model, _) => Browser.view(model, address)
```
Gozala: We use this hack of wrapping address so that we could delegate every action to a services that were implemented in the past. In a future most services will be replaced by effects while some of the use cases will be replaced by forwarding addresses.

Gozala:
> Loading a new tab causes frame drop and runs for more than 150ms that means something is either wrong with timing or something is going wrong there. Unfortunately profiler seems broken in devtools but I'll investigate what is going on there.

Ok I have figured out what is going on there posting an image that illustrates:

screen shot 2015-10-26 at 3 47 57 pm

Turns out it's gecko takes 200+ms to insert a mozbrowser iframe into document, which does explains while we get frame drop warning.

gordonbrander:
I've tagged current master at `0.0.10`. I'm going to merge this now so we can build the new UI in master on top of the new code base.

PR #684, opened 2015-10-30
Perspective UI prototype

Gozala:
For now I primarily added `type` directory that encodes details present in the specification diagram into type signatures. From this point on I intend to build up a UI that fulfills presented types signatures.


type/browser/shell.js
```diff
@@ -0,0 +1,12 @@
+/* @flow */
+
+import * as Focusable from "../common/focusable"
+import * as Hoverable from "../common/focusable"
```
Gozala: should be '../common/hoverable'

type/common/focusable.js
```diff
@@ -0,0 +1,30 @@
...
+
+export type Action
+  = RequestFocus
+  | Focus
+  | Blur
+
+type Focusable = Model
+
+
+export type focus <Model:Focusable> = (model:Model) => Model
```
Gozala:
```js
type X = { isX: true } & Focusable
type Y = { isY: true } & Focusable

type foo = (model:Focusable) => Focusable
```

type/browser/input.js
```diff
@@ -0,0 +1,13 @@
...
+import * as Focusable from "../common/focusable"
+
+export type Model
+  = Editable.Model
+  & Focusable.Model
+
+
+export type Action
+  = Editable.Action
+  & Focusable.Action
```
Gozala: should be `|` instead of `&`

type/browser/browser.js
```diff
@@ -0,0 +1,38 @@
...
+  | For<"Assistant", Assistant.Action>
+  | For<"Devtools", Devtools.Action>
+
+export type Response
+  = For<"WebViews", WebViewList.Action>
+  | For<"Assistant", Assistant.Response>
+  | For<"Devtools", Devtools.Response>
+
+export type view = (model:Model, address:Address<Action>) => VirtualTree
+export type step = (model:Model, action:Action) => [Model, Effects<Response>]
```
Gozala:
```js
const asFor = target => action => ({type: "For", target, action})
const merge = (target, extra) => Object.assign({}, target, extra)

view = (model, address) =>
  html.div({
    key: 'root',
    title: //...,
    onFocus: onWindow(forward(address, asFor("Shell")), Focusable.asFocus),
    onBlur: onWindow(forward(address, asFor("Blur")), Focusable.asBlur),
  }, [
    WindowControls.view(...),
    Devtools.view(model.devtools, forward(address, asFor("Devtools"))),
    // ...
  ])


step = (model, action) => {
  if (action.type === "For") {
    if (action.target === "Shell") {
      return [
        merge(model, {
          shell: Shell.update(model.shell, action.action)
        }),
        Effects.none
      ]
    }

     if (action.target === "Devtools") {
       const [devtools, fx] = Devtools.step(model.devtools, action.action)
       return [
         merge(model, {devtools}),
         fx.map(asFor("Devtools"))
       ]
     }
  }
}
```

gordonbrander:
We previously had higher-order functions like `on` and `onWindow` https://github.com/mozilla/browser.html/blob/1e04184e149a299df2cc8008dc1433ed6a04120e/src/driver/virtual-dom/index.js#L120. These seem to have disappeared.

gordonbrander:
Do we use immutable.js at all in this new version? It doesn't look like it.

gordonbrander:
FYI: I'm attempting to build on top of this PR here https://github.com/Gozala/browser.html/pull/2/files

gordonbrander:
Updated PR here https://github.com/mozilla/browser.html/pull/768

PR #696, opened 2015-11-19
Add mode field to application model

gordonbrander:


gordonbrander:
Taking a different approach.

PR #697, opened 2015-11-19
Perspective UI modal views

gordonbrander:


type/browser/browser.js
```diff
@@ -13,6 +14,7 @@ import * as Assistant from "./assistant"
 
 export type Model = {
   version: Version,
+  mode: PerspectiveUI.Model,
```
Gozala: Browser module should not be aware of perspective UI at all perspective UI should embed browser instead IMO.

gordonbrander: Ok, sounds good, but I'm unclear on what that should look like.

src/browser/sidebar.js
```diff
@@ -107,3 +108,13 @@ export const view = ({entries}, address) =>
       className: 'sidebar-toolbar'
     })
   ]);
+
```
Gozala: I do think that embedder of side bar should just wrap it in a container that is displayed or not based on the mode off the embedder. Dose that make sense ?

gordonbrander: @Gozala that's not going to work when we start animating between modes. Transitions are more complex than just hide/show. We need to be able to modify the style of the root view as it transitions.

PR #699, opened 2015-11-19
Stab out mode switching code.

Gozala:


src/browser/perspective-ui.js
```diff
@@ -0,0 +1,162 @@
...
+import {Effects, html, forward, thunk} from "reflex";
+
+import * as Shell from "./shell";
+import * as Input from "./input";
+import * as Assistant from "./assistant";
+import * as WindowControls from "./window-controls";
+import * as Sidebar from "./sidebar";
+
+
+const initialize/*:type.initialize*/ = () => {
```
gordonbrander: We should be exporting all of these functions :)

src/browser/perspective-ui.js
```diff
@@ -0,0 +1,159 @@
...
+      if (model.browser.webViews.entries.length > 0) {
+        return [asShowWebView(browser), fx];
+      }
+    }
+    else if (isSubmit(action)) {
+      const [browser, fx] = Browser.step(model.browser, action);
+      return [asShowWebView(browser), fx];
+    }
+    // @TODO: Probably we should prevent input field from loosing a focus
+    // by sendig in FocusRequest on Blur. Also we may want to ignore tab
```
gordonbrander: We should switch to `select-web-view` in this case.

src/browser/perspective-ui.js
```diff
@@ -0,0 +1,159 @@
...
+      // to do.
+      if (model.browser.webViews.entries.length > 0) {
+        return [asShowWebView(browser), fx];
+      }
+    }
+    else if (isSubmit(action)) {
+      const [browser, fx] = Browser.step(model.browser, action);
+      return [asShowWebView(browser), fx];
+    }
+    // @TODO: Probably we should prevent input field from loosing a focus
```
gordonbrander: Yeah, let's keep the input focused in this scenario.

src/browser/perspective-ui.js
```diff
@@ -0,0 +1,159 @@
...
+        isActivateWebView(action))
+    {
+      const [browser, fx] = Browser.step(model.browser, action);
+      return [asShowWebView(browser), fx];
+    }
+    else if (isCreateTab(action)) {
+      const [browser, fx] = Browser.step(model.browser, action);
+      return [asCreateWebView(browser), fx];
+    }
+    // @TODO: Find out if we should handle this as it's not in the control
```
gordonbrander: Yes, `meta + l` should still switch to `editWebView` (the one currently active)

PR #712, opened 2015-11-21
3D transform webview in select and show-tabs modes

gordonbrander:


src/browser/perspective-ui.js
```diff
@@ -180,7 +180,7 @@ export const step = (model, action) => {
   return [merge(model, {browser}), fx];
 }
 
-const styles = StyleSheet.create({
+const style = StyleSheet.create({
```
gordonbrander: Naming consistency change.

PR #714, opened 2015-11-21
Send focus action on click (please review)

gordonbrander:
Incomplete. Don't merge.

### DIFF
--- a/src/browser/perspective-ui.js
+++ b/src/browser/perspective-ui.js
@@ -78,6 +78,10 @@ export const isActivateWebView = action =>
   isWebViewAction(action) &&
   action.action.action.type === 'WebView.Activate';
 
+export const isEditWebview = action =>
+  isWebViewAction(action) &&
+  action.action.action.type === 'WebView.Edit';
+
 export const isSelectRelativeWebView = action =>
   action.type === 'For' &&
   action.target === 'webViews' &&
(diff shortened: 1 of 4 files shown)


src/browser/web-view.js
```diff
@@ -341,6 +342,8 @@ const viewFrame = (model, address) =>
     onMozBrowserScrollAreaChanged: on(address, decodeScrollAreaChange),
   });
 
+const asForInput = asFor('input');
```
gordonbrander: @TODO wrapping action in `For` here means that by the time the action is sent, it is triple-wrapped. Is that what we want?

src/browser/perspective-ui.js
```diff
@@ -78,6 +78,10 @@ export const isActivateWebView = action =>
   isWebViewAction(action) &&
   action.action.action.type === 'WebView.Activate';
 
+export const isEditWebview = action =>
```
gordonbrander: Should these types of functions live in a common file?

src/browser/perspective-ui.js
```diff
@@ -130,6 +134,10 @@ export const step = (model, action) => {
       const [browser, fx] = Browser.step(model.browser, action);
       return [asShowWebView(browser), fx];
     }
+    else if (action.target === 'Browser.KeyUp' && action.action.key === 'Escape') {
```
gordonbrander: Is this how we want to do this, or should I be using https://github.com/mozilla/browser.html/pull/714/files#diff-5c204f4fed6072a5044a37de83d15202R60? Except that `isEscape` doesn't seem to be defined or used anywhere. Is it actually a predicate function for detecting escape key?

PR #732, opened 2015-11-25
Use documentElement for global events instead of window

Gozala:


Gozala:
cc @paulrouget

paulrouget:
(thanks)

PR #733, opened 2015-11-25
Use keyup event instead of input event for observing input updates.

Gozala:
@paulrouget use of keyup event seems to have some issues when typing fast (one you mentioned in the meeting) but even on gecko. I get an impression that when `keyup` is fired `input.value` has not yet changed, but it maybe something different. I'll try to work it out what happens tomorrow, but for now you can try merging this for testing on servo if you like.


gordonbrander:
Tested this against latest perspective UI, and it works. Merging as it fixes a major UI issue in Servo.

PR #748, opened 2015-11-27
Hide tabs animation

Gozala:


Gozala:
Made this by mistake!

PR #749, opened 2015-11-27
Hook hide tab transition & switch tab transitions.

Gozala:


src/browser/perspective-ui.js
```diff
@@ -389,6 +422,21 @@ const transition = {
     return {
       transform: `translateX(${x}px) rotateY(${angle}deg)`
     };
+  },
+  sidebarHide(model) {
+    const hidden = {angle: -15, x: 420};
```
Gozala: @gordonbrander In your changes to `sidebarShow` in starting (side bar hidden) state `angle` is `-45` and in such state sidebar is still visible and overlays web-views. There for in `sidebarHide` I'm using `-15` instead so that in hidden state side bar will not b visible. Note you did not see sidebar in your changes because in `show-web-view` mode view function did not used angle at all.

P.S.: ideally `hidden` state and `shown` state of the side bar will be same value across animations.

PR #763, opened 2015-12-04
Fix window dragging issues.

Gozala:


gordonbrander:
Works. (Though in future, we may want to restrict draggable area to title bar area)

PR #764, opened 2015-12-04
Sliding tab switcher - Do not merge!

Gozala:


Gozala:
We can always find this pull if we need to, so closing it for now.

PR #779, opened 2015-12-14
Decouple overlay from the perspective-ui.

Gozala:


src/browser/overlay.js
```diff
@@ -16,80 +17,116 @@ const visible/*:type.Visible*/ = 0.1;
...
+  });
+
+export const Click = always({type: "Click"});
+export const Show = {type: "Show"};
+export const Hide = {type: "Hide"};
+export const Fade = {type: "Fade"};
+
+const Animation = action => ({type: "Animation", action});
+const Showed = always({type: "Showed"});
+const Hided = always({type: "Hided"});
```
gordonbrander: This should be "hidden"

src/browser/overlay.js
```diff
@@ -16,80 +17,116 @@ const visible/*:type.Visible*/ = 0.1;
...
+      : {opacity: invisible}
+  });
+
+export const Click = always({type: "Click"});
+export const Show = {type: "Show"};
+export const Hide = {type: "Hide"};
+export const Fade = {type: "Fade"};
+
+const Animation = action => ({type: "Animation", action});
+const Showed = always({type: "Showed"});
```
gordonbrander: This should be "Shown"

PR #795, opened 2016-01-06
Refactor code base to follow design architecture

Gozala:
This pull request aims to do following:

1. [x] Get rid of nested else if blocks and instruction repetitions (like you can find in `perspective-ui.js`).
2. [x] Get rid of the single mode transition animation that drives all of the animations observed in UI during transition.
3. [x] Adds missing type information.
4. [x] Reduces number of actions being dispatched, by only dispatching actions bound to they keybindings. In other words keybindings are no longer produce actions unless there is a matching binding for it.
5. [x] No longer have views per mode that slightly alter layout and pass context styles. Instead parent modules orchestrate mode transitions for children by updating them with an action that switches child into specific mode.
6. [x] Address passed to children now is forwarded via annotator function that:
  1. Translates actions it needs to react to into own actions actions. That way update never needs to unpack received actions.
  2. Boxes actions that just need to be delegated to it's originating child.

  **Note:** Child effects are also mapped with a same annotation function.
7. [x] Few general components have being added, and existing components like window controls have being refactored to leverage those.


Gozala:
@gordonbrander BTW I found few issues with our code when I was doing this work. There is one I'd need your help to resolve it - When we enter the full screen mode toggle (green) button should have a different image when hovered. Would be really cool if you could add it to the current button `window-controls.sprite.png`.

PR #830, opened 2016-01-15
Fix input styles

gordonbrander:
Fixes https://github.com/mozilla/browser.html/issues/826.

- Move icon padding from input wrapper to input. This won't fix the dotted focus line in Servo, but it will at least make it wrap around the input instead of a random box.
- Add focus border to input.

### DIFF
--- a/src/browser/input.js
+++ b/src/browser/input.js
@@ -155,31 +155,36 @@ const readSelect = compose
       Editable.Select(readSelection(target))
   );
 
-const inputWidth = '460px';
-const inputHeight = '40px';
+const inputWidth = 460;
+const inputHeight = 40;
+const inputXPadding = 32;
 
 const style = StyleSheet.create({
   combobox: {
-    background: '#EBEEF2',
-    borderRadius: '5px',
     height: inputHeight,
     left: '50%',
-    marginLeft: `calc(-1 * (${inputWidth} / 2))`,
+    marginLeft: `calc(-1 * (${inputWidth}px / 2))`,
     position: 'absolute',
     top: '40px',
-    width: inputWidth
+    width: `${inputWidth}px`
   },
   field: {
-    background: 'transparent',
-    borderWidth: 0,
+    background: '#EBEEF2',
+    borderRadius: '5px',
+    borderWidth: '3px',
+    borderStyle: 'solid',
+    borderColor: 'transparent',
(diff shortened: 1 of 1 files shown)


src/browser/input.js
```diff
@@ -219,9 +220,7 @@ export const view/*:type.view*/ = (model, address) =>
   html.div({
     className: 'input-combobox',
     style: Style( style.combobox
-                ,   model.isVisible
-                  ? style.visible
-                  : style.hidden
+                , !model.isVisible && style.hidden
```
Gozala: This is necessary change.

src/browser/input.js
```diff
@@ -155,32 +155,36 @@ const readSelect = compose
...
     MozAppearance: 'none',
-    height: inputHeight,
-    lineHeight: inputHeight,
+    height: `${inputHeight - 6}px`,
+    lineHeight: `${inputHeight - 6}px`,
     margin: 0,
-    padding: 0,
-    width: `calc(${inputWidth} - ${32 * 2}px)`
+    padding: `0 ${inputXPadding}px`,
+    width: `calc(${inputWidth - 6}px - ${inputXPadding * 2}px)`
```
Gozala: Seem to me that we don't need calc as we can calculate that directly.

src/browser/input.js
```diff
@@ -155,32 +155,36 @@ const readSelect = compose
...
+const inputXPadding = 32;
 
 const style = StyleSheet.create({
   combobox: {
-    background: '#EBEEF2',
-    borderRadius: '5px',
     height: inputHeight,
     left: '50%',
-    marginLeft: `calc(-1 * (${inputWidth} / 2))`,
+    marginLeft: `calc(-1 * (${inputWidth}px / 2))`,
```
Gozala: Seems like we don't need a calc here either as we can calculate it directly.

src/browser/input.js
gordonbrander: Oops, you are right. This is a mistake left over from the previous implementation.

src/browser/input.js
gordonbrander: Ditto.

src/browser/input.js
gordonbrander: No, just snuck it in because I was already in the file.

PR #850, opened 2016-01-26
Do not select tab on close button click.

### LINKED ISSUE #849: Clicking close button on tabs sometimes doesn't work
It's very hard since the hit area is very small.

Gozala:
Fixes #849 

Clicking tab close button was triggering `onMouseDown` of the tab container triggering `TabSelect` action that was starting an animation & as a consequence `onClick` on close button was no longer triggered which cased unreliable tab closing behavior.

As a matter of fact `Select` action was not required and was more of leftover from post refactoring codebase.

### DIFF
--- a/src/browser/browser.js
+++ b/src/browser/browser.js
@@ -86,8 +86,6 @@ const SidebarAction = action =>
   ? CreateWebView
   : action.type === "ActivateTab"
   ? ActivateWebViewByID(action.id)
-  : action.type === "SelectTab"
-  ? SelectWebViewByID(action.id)
   : action.type === "CloseTab"
   ? CloseWebViewByID(action.id)
   : action.type === "Tabs"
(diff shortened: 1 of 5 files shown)


gordonbrander:
R+. Merging.

There is an additional problem: when closing a tab in sidebar, tab is activated and mode zooms in to `show-webview`. This shouldn't happen (tab should simply close). However, we can address this in a follow-up fix.

Gozala:
> There is an additional problem: when closing a tab in sidebar, tab is activated and mode zooms in to show-webview. This shouldn't happen (tab should simply close). However, we can address this in a follow-up fix.

Right I forgot to post a comment about it in the issues, the problem is that `onMouseUp` is still triggered causing `Activate` action to occur. Unfortunately there is no obvious & simple way to prevent that propagation, but that's what I think we should do.

PR #853, opened 2016-01-26
Stop propagation of events in close button

### LINKED ISSUE #852: Closing a tab should not zoom in webview
Follow-up from https://github.com/mozilla/browser.html/pull/850#issuecomment-174782325.

When closing a tab in sidebar, tab is activated and mode zooms in to show-webview. This shouldn't happen (tab should simply close).

Gozala:
This prevents tab from triggering TabActivate action when tab’s close button is clicked.

Fixes #852


gordonbrander:
R+. Works great.

PR #863, opened 2016-02-05
Search suggestions

Gozala:
Initial patch will just add search suggestions.


src/browser/assistant/search.js
```diff
@@ -0,0 +1,308 @@
...
+    if (pendingRequests[id] != null) {
+      pendingRequests[id].abort();
+      delete pendingRequests[id];
+    }
+  })
+
+const search/*:Search.search*/ =
+  (id, input, limit) =>
+  Task.future(() => new Promise(resolve => {
+    const request = new XMLHttpRequest({ mozSystem: true });
```
gordonbrander: Nit: we should have a comment that explains we use XHR instead of `fetch` because we need to set `mozSystem: true`.

PR #864, opened 2016-02-06
Enter create-tap page once last tab is closed.

Gozala:
I think current behavior of entering screen with no web-view is kind of odd, so this patch just enters `create-web-view` mode if last tab is closed. 

@gordonbrander Feel free to land or tell me that we should do something else here.


gordonbrander:
Awesome! This PR is indeed how it's supposed to work. (I thought this case was in the state diagram, but I might have missed it?)

gordonbrander:
It works! R+.

PR #868, opened 2016-02-09
Fix crossfade in selection transition

### LINKED ISSUE #865: Background shouldn't show through Webview crossfade
I think currently both webviews use a fade animation. Instead, it should work like this:

1. Current webview is is set to `z-index: 2`. Next webview is set to `z-index: 1; opacity: 1;`
2. Current webview fade animation begins. Because next webview is behind, you get  a clean crossfade.
3. When animation is complete, current webview `z-index: 1`, next webview `z-index: 2`.

gordonbrander:
Fixes #865

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -423,7 +423,7 @@ const selected = model =>
 
 const unselect = model =>
   ( model.isSelected
-  ? startUnselectAnimation(merge(model, {isSelected: false}))
+  ? [ merge(model, {isSelected: false}), Effects.none ]
   : [ model, Effects.none ]
   );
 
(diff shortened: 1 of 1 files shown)


gordonbrander:
Gozala gives it an R+. https://browserhtml.slack.com/archives/general/p1455042029000655

src/browser/web-view.js
```diff
@@ -221,18 +221,6 @@ const SelectAnimationAction = action =>
     }
   );
 
-const UnselectAnimationAction = action =>
-  ( action.type === "End"
-  ? Unselected
```
Gozala: I just noticed we no longer trigger `Unselected` action which parent component may be expecting (can't remember right now).

Issue #854, opened 2016-01-26
disable logs by default


paulrouget:
I don't care much about the logs while working on Servo. Can we make them optional? Maybe behind one the devtools checkbox?

Gozala:
@paulrouget In fact I already have the patch that only enables them if you have certain setting set. Problem is we don't have settings in servo though :)

Feel free to remove it for now and I can put it back once I'll get around to polyfilling settings API.

paulrouget:
Should I just comment out the calls to `console.log`?

PR #877, opened 2016-02-22
remove logs

paulrouget:
Fix https://github.com/browserhtml/browserhtml/issues/854


Gozala:
This is made obsolete by @gordonbrander #879

Issue #875, opened 2016-02-18
Make it possible to pass options via the hash


paulrouget:
For example: http://localhost:6060/index.html#window_controls=false;devtools=false

paulrouget:
It would help a lot if we could pass at least 2 options:
1. window controls (enabled or not)
2. a list of tabs to open at startup

paulrouget:
Add to that add:
3. enable / disable logs

paulrouget:
… or add a `option.js` file next to `index.html` where I can set some variables.

paulrouget:
I basically want to be able to share a state with other servo developers. Whatever the solution is (external file or hash)

PR #879, opened 2016-02-22
Add config file for logging and devtools

gordonbrander:
This is a simple commit that adds a `browserhtml.json` config file for developers. This file lets us configure default settings.

Fixes https://github.com/browserhtml/browserhtml/issues/875.

@Gozala do you have strong opinions about this? It's the simplest possible solution I could think of.

### DIFF
--- a/browserhtml.json
+++ b/browserhtml.json
@@ -1,0 +1,3 @@
+{ "logging": true
+, "devtools": false
+}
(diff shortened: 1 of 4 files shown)


paulrouget:
Can you make the logging false by default?

Also, name this file `default-prefs.json`.

We should pull a file named `custom-prefs.json` or just `prefs.json`, that should be optional (absent by default). Then someone can add its own configuration without polluting its tree.

Being able to add comments to such a file will be important. JSON doesn't support comments. So maybe we can use something like that: https://github.com/sindresorhus/strip-json-comments

PR #885, opened 2016-02-25
WIP about:newtab

gordonbrander:
WORK IN PROGRESS. DO NOT MERGE.


gordonbrander:
Nope. This is the right way to do it, but trying a simpler approach first that doesn't require refactoring how tabs open/close.

PR #892, opened 2016-03-01
Add Cargo files

metajack:
I forgot to git add these in the previous PR.


Gozala:
End up doing #901 instead.

PR #896, opened 2016-03-02
Fix the favicon fallback handling.

### LINKED ISSUE #897: Favicon fallback doesn't work as expected
For example following page http://codemirror.net/mode/elm/index.html has favicon in other browsers but it does not in browser.html

Gozala:
Fixes #897


Gozala:
@gordonbrander can you please take a look at this ?

gordonbrander:
R+

PR #904, opened 2016-03-03
Resolve the issue with forced input replacement

### LINKED ISSUE #869: Search suggestions are forced, blowing away what you type
There's a bug with the current search implementation that blows away what you type, replacing it with the current search suggestion. This means you can't actually type what you want if it doesn't match the first suggestion.

Screen recording: https://drive.google.com/file/d/0B1az_KF0ZYc_aGh1cHdwVVZoTHc/view?usp=sharing

cc @Gozala

Gozala:
Fixes #869

This fix is more of a hack than a proper fix, but it does make input usable.

### DIFF
--- a/src/browser/assistant/search.js
+++ b/src/browser/assistant/search.js
@@ -211,6 +211,14 @@ const updateMatches = (model, result) =>
   );
 
 const replaceMatches = (model, results) => {
+  const top = results[0]
+  const query = model.query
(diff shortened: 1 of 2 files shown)


Gozala:
@gordonbrander can you please take a look ?

Gozala:
@gordonbrander said merge it!

Issue #911, opened 2016-03-04
Remove the udpate code


paulrouget:
Apparently, the update will come from Servo itself or from the mozilla updater.
We don't have a strategy for web updates yet, and we don't have any security model.

Let's remove the update code.

PR #912, opened 2016-03-04
remove udpater

paulrouget:


src/browser/browser.js
```diff
@@ -731,8 +652,6 @@ export const update/*:type.update*/ = (model, action) =>
   ? expand(model)
   : action.type === "Shrink"
   ? shrink(model)
-  : action.type === "ResizeAnimation"
```
gordonbrander: This should not be removed.

src/browser/browser.js
```diff
@@ -641,59 +615,6 @@ const endResizeAnimation = model => {
 const shrinked = endResizeAnimation;
 const expanded = endResizeAnimation;
 
-const updateResizeAnimation = (model, action) => {
```
gordonbrander: This should not be removed.

paulrouget:
Fix https://github.com/browserhtml/browserhtml/issues/911

Issue #878, opened 2016-02-22
Update browserhtml/docs/


gordonbrander:
`architecture.md` and `Synthesis UI.md` are out-of-date. It might be better to remove them and link to a wiki page from `README.md`.

@Gozala any preference?

Gozala:
Sure remove them, I'll try to get them up to date after probably after / or maybe by June.

PR #920, opened 2016-03-08
Fix "incerement" misspelling.

notriddle:
Spelling fix in architecture.md.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -58,7 +58,7 @@ export const update = (model, action) =>
   model;
 ```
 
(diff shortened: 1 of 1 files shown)


gordonbrander:
Thanks for this. Also note the architecture docs need to be updated a bit https://github.com/browserhtml/browserhtml/issues/878. We've improved a few things for performance reasons.

The basic gist is the same, but we ditched Immutable.js for ordinary JavaScript objects. We're borrowing patterns from https://github.com/evancz/elm-architecture-tutorial.

Issue #915, opened 2016-03-07
Add window controls only if titlebar-less window


paulrouget:
We need window controls only on Mac, when the `-b` option is passed.

I suggest that we pass options to the document via the hash to tell browerhtml if it should use window controls or not.

For example: `index.html#titlebar=true`

paulrouget:
@gordonbrander / @Gozala can one of you look at this?

paulrouget:
I'll take care of it.

paulrouget:
A easier solution is probably to just use `navigator.platform` and just assume no native titlebar on mac.

gordonbrander:
@paulrouget why not use `browserhtml.json` config file? We could allow `.mach` to choose a config file, for example.

paulrouget:
> @paulrouget why not use browserhtml.json config file? We could allow .mach/ to specify a config file, for example.

So when we package Servo, we would put a `os:…` in `browserhtml.json`?
That won't help. Instead of querying the OS, we would query the json file.

We need a way for the runtime, at *runtime*, to tell if a native titlebar is used or not. This will happen once we get the proper Electron-like runtime.

gordonbrander:
@paulrouget we're creating different app packages for the different OS's. This means we can have different config files for these different app packages. Seems like a simple fix, but I'm not married to it. I'm open to whatever solution makes sense.

Issue #925, opened 2016-03-14
Make it look good on Windows


paulrouget:
Not sure yet what we want to do here, but browserhtml needs to work correctly on Windows.

paulrouget:
This [comment](https://github.com/browserhtml/browserhtml/issues/913#issuecomment-197912816) applies here as well.

paulrouget:
I'm about to fix #915. @gordonbrander Is there anything we want to do here?

gordonbrander:
Do we hide the title bar on windows?

paulrouget:
> @paulrouget do we hide the title bar on windows?

titlebar will be native.

paulrouget:
I consider this done.

Issue #913, opened 2016-03-04
Make it look good on Linux


paulrouget:
Not sure yet what we want to do here, but browserhtml needs to work correctly on Linux.

Anachron:
I have ArchLinux at home, so I could be testing and debugging.

paulrouget:
I think there are 2 things we can do:

Option 1, figure out a way to make a border-less draggable and resizable window on Linux. I guess that's only possible if we use a widget toolkit, or if we use wayland (which is supported),

Option 2,  we just rely on the window manager to draw the window decoration (as it is now), but then we need to hide the window controls (osx buttons for now).

I would go for (2) first.

This means browser.html needs to know if decorations/titlebar surround the servo window. And I would suggest to do this: https://github.com/browserhtml/browserhtml/issues/915

Anachron:
I'd go for 2 as well. I have 3 different wms for testing 1 though. 

I'd not recommend Wayland yet. Even though its testable, it still segfaults for me. 

When will you get to check #915?

paulrouget:
> When will you get to check #915?

Not anytime soon.

I do have a Linux box here. The only thing that stands out is the titlebar.

paulrouget:
I'm about to fix #915. @gordonbrander Is there anything we want to do here?

gordonbrander:
@paulrouget we will not hide the title bar on Linux, correct?

paulrouget:
> @paulrouget we will not hide the title bar on Linux, correct?

titlebar will be native.

paulrouget:
I consider this done.

PR #942, opened 2016-03-21
Only show controls on Mac

paulrouget:
Closes: #915 #925 #913

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -39,7 +39,17 @@ body {
   font-size: 12px;
   line-height: 1.5;
   overflow: hidden;
+}
+
+/* We want the native window to have round borders. When a
+ native titlebar is used, the round borders come from the
+ OS itself */
+body:not(.use-native-titlebar) {
   border-radius: 6px;
+}
+
+body.use-native-titlebar {
+  border-radius: 0;
 }
 
 .webview-combobox:hover {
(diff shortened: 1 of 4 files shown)


css/theme.css
```diff
@@ -42,6 +42,10 @@ body {
   border-radius: 6px;
 }
 
+body.use-native-titlebar {
+  border-radius: 0;
+}
```
Gozala: @paulrouget Can you write a comment explaining the story about the border-radius here ? I have no idea why it has being there in first place or why is it that we don't want it if native titlebar is used.

src/browser/index.js
```diff
@@ -39,6 +40,9 @@ if (isReload) {
   window.application.address(UI.LiveReload);
 }
 
+document.body.classList.toggle('use-native-titlebar',
+                               Runtime.useNativeTitlebar());
+
```
Gozala: I wonder if we should just add `OS.platform()` to the `ClassList` instead.

paulrouget: I'd prefer to keep the logic that tells if native titlebar is used or not, isolated.

Eventually, the runtime will tell us what kind of titlebar is used, and we will just need to change what's within runtime.js.

css/theme.css
paulrouget: ok

paulrouget:
Added a comment.

Issue #922, opened 2016-03-11
remove setSelectionRange test


paulrouget:
… it's now implemented in Servo

Issue #914, opened 2016-03-07
Re-use vw and vh


paulrouget:
These should be working again very soon (once I get a new build).

paulrouget:
https://github.com/servo/servo/issues/8754

PR #943, opened 2016-03-21
remove some Servo workarounds

paulrouget:


paulrouget:
Fixes https://github.com/browserhtml/browserhtml/issues/922 and https://github.com/browserhtml/browserhtml/issues/914

Gozala:
@paulrouget I wonder if change from `100%` to `100vh` / `100vw` is really buying us anything here ? @mbrubeck also said it maybe slow. 

Either way I'll leave it up to you. Changes look good though so fell free to land as is or without changes that I've commented on above.

paulrouget:
Using `vh` will force overflow if we do any layout mistake. I'd prefer to go that way.

Issue #872, opened 2016-02-16
Integrate force touch for switching to show-tabs mode


gordonbrander:
Wireframe forthcoming.

paulrouget:
For the record, there are 4 phases:

1. touchpad at rest (`force = 0`)
2. pressure on the touchpad, first click not reached (`0 < force < 1`)
3. pressure on the touchpad, first click reached, second click not reached (force: ``1 < force < 2``)
4. pressure on the touchpad, second click reached (force: `2 < force < 3`)

Clicking happens when `force = 1` and `force = 2`.

paulrouget:
Here is a quick-and-dirty hack that shows how it can be done:

```js
var THRESHOLD = 1.5;
var MAX = 2.5;

window.addEventListener("load", () => {
  var stack = document.querySelector(".webviews-stack");
  var overlay = document.querySelector(".overlay");
  var sidebar = document.querySelector(".sidebar");

  function update(t) {
    stack.style.transform = `translate3d(0px, 0px, ${-200 * t}px)`;
    overlay.style.opacity = 0.1 * t;
    sidebar.style.transform = `translateX(${(1 - t) * 550}px)`;
  }

  document.documentElement.addEventListener("webkitmouseforcechanged", e => {
    var force = e.webkitForce;
    if (force >= THRESHOLD && force <= MAX) {
      update((force - THRESHOLD) / (MAX - THRESHOLD));
    } else if (force < THRESHOLD) {
      update(0);
    }
  });
  document.documentElement.addEventListener("mouseup", e => {
    update(0);
  });
});
```

paulrouget:
demo: https://streamable.com/5j47

paulrouget:
I'm working a proper PR for Servo. I do not use the `webkit` prefix anymore, but `servo`.

paulrouget:
Any update?

gordonbrander:
Working on it now :)

gordonbrander:
Note to self: use `--pref dom.forcetouch.enabled` to toggle force touch on in Servo.

PR #946, opened 2016-03-22
Transition to ShowTabs on force touch

gordonbrander:
Fixes https://github.com/browserhtml/browserhtml/issues/872.

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -84,6 +84,20 @@ export const Focus/*:type.Focus*/ =
   { type: 'Focus'
   };
 
+// The push action is sent by Force Touch on webview.
+export const Push = force =>
+  ( { type: 'Push'
+    , force
+    }
+  );
+
+// This action is sent up as effect to WebViews.
+export const Pushed = force =>
+  ( { type: 'Pushed'
+    , force
+    }
+  );
+
 export const Load/*:type.Load*/ = uri =>
   ( { type: 'Load'
     , uri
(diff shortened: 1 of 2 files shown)


paulrouget:
- For some reason, this is very very slow on my computer. Any idea why? Do we spawn a lot of parallel animation tasks every time pressure is greater than the threshold
- s/webkit/servo/ - please update to the latest version of servo (and use `--pref dom.forcetouch.enabled`)
- this should include a "cancellable zone". This is what happens with other forceTouch feature in OSX:
 - let's call `z` a pressure between `1` and `2`
 - between `0` and `z`, do nothing
 - between `z` and `2`, animate based it on the pressure value (releasing slightly should run the animation back a little, pressing slightly hard should run the animation forward a little). Releasing the touchpad cancel the animation
 - `> 2`, finished the whole animation

gordonbrander:
> s/webkit/servo/ - please update to the latest version of servo (and use --pref dom.forcetouch.enabled)

Fixed.

> For some reason, this is very very slow on my computer. Any idea why? Do we spawn a lot of parallel animation tasks every time pressure is greater than the threshold.

Nothing crazy happening, to my knowledge. We send some small objects for each push event, but we only animate once, for the first push event above the threshold. Will investigate.

> this should include a "cancellable zone". This is what happens with other forceTouch feature in OSX

I tried this first. However there was not enough "depth" to force touch for this to make a difference to the user experience. My recommendation would be that we figure out what the perf issue is, then land this, then add nuance to it in a follow-up commit.

paulrouget:
> However there was not enough "depth" to force touch for this to make a difference to the user experience

That all depends on your settings. If you allow a firmer click, you do have the time to cancel. Let's follow cocoa's behavior:

1) do not use an arbitrary threshold to trigger the sidebar, do it on `servomouseforcedown`
2) build the transitional stage (between 1 and 2)

This is how any other forceTouch feature works.

PR #947, opened 2016-03-22
change iframe background color

### LINKED ISSUE #934: The overscroll background color doesn't mach the titlebar color
screen shot 2016-03-16 at 09 40 10

paulrouget:
Overscroll shows the background color of the iframe.
We want this to match the titlebar color.

Fix #934

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -726,7 +726,12 @@ const viewFrame = (model, address) =>
     'data-name': model.name,
     'data-features': model.features,
     element: Driver.element,
-    style: styleSheet.iframe,
+    style: Style(styleSheet.iframe
+                , ( model.page.pallet.background != null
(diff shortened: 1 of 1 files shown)


Gozala:
> Overscroll shows the background color of the iframe.

If iframe has no background is it going to to be background color of the element behind it or ? If so, I think it may make more sense to just set background on the web-view div itself and not titlebar and iframe individually.

Either way this seems legit so feel free to land.

paulrouget:
> If iframe has no background is it going to to be background color of the element behind it or ?

Yes.

> If so, I think it may make more sense to just set background on the web-view div itself and not titlebar and iframe individually.

I'd prefer to never have transparent iframes. In theory, that should not be a problem, but I'm not comfortable with non-content elements to potentially show up.

PR #948, opened 2016-03-22
update locationchange event to work both with Servo and Gecko

paulrouget:


Gozala:
Looks legit. Thanks @paulrouget

PR #958, opened 2016-03-25
Handle window close/reload diff on servo

### LINKED ISSUE #944: If Servo, use window methods instead of mozContentEvent
In runtime.js, until we get a proper runtime, can we use:
- `window.close()` for `Runtime.quit`
- `window.location.reload()` for: `Runtime.reload`
- `window.location.reload(true)` for: `Runtime.cleanReload`

I'm not sure about fullscreen and minimize yet.

@Gozala can you look at this? The Task thing in `runtime.js` is confusing me.

Gozala:
Runtime.quit -> window.close()
Runtime.reload -> window.location.reload()
Runtime.cleanReload -> Runtime.location.reload(true)

Fixes #944 Depends on #957

### DIFF
--- a/src/common/runtime.js
+++ b/src/common/runtime.js
@@ -63,6 +63,12 @@ export const DownloadUpdate =
     }
   );
 
+export const isServo/*:boolean*/ =
+  window
+  .navigator
+  .userAgent
+  .toLowerCase()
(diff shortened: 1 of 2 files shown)


paulrouget:
Thank you!

PR #959, opened 2016-03-25
Make deployment script incremental

Gozala:
Deploying new version no longer wipes out previously deployed version. Fixes servo/servo#10170


gordonbrander:
R+

PR #960, opened 2016-03-25
Add babel-polyfill to devDependencies

Gozala:
This fixes regression I introduced #957, where I forgot to include babel-polyfill into dependencies.


gordonbrander:
R+

PR #961, opened 2016-03-25
Nothing to review, just trying to verify that travis fails builds when types mismatch

Gozala:


Gozala:
So failures are caught by flow and reported by travis, my only guess is somehow `babel-polyfill` is available in travis.ci

Gozala:
Alright I gave up investigating why it did not failed on missing `babel-polyfill` does not seem like it's worth spending too much time on this given that builds do fail on missing dependencies or other errors. I just assume that travis environment has `babel-polyfill` and there for  `flow` is able to find it.

PR #962, opened 2016-03-25
Fix brief appearance of full length progress bar.

### LINKED ISSUE #928: Before the progressbar starts growing, it's displayed full size for a very short time

Gozala:
Before `display.x` was changing from 1 to 0 and when progress idle `display.x` was set to 0 which meant seeing full progress bar. This change reverses meaning of x so it grows from 0 to 1.

Fixes #928


gordonbrander:
Works! R+.

Issue #941, opened 2016-03-21
Only show the back button if we can go back


paulrouget:
Or maybe grey it out? @gordonbrander what do you think?

`getCanGoBack` is not implemented in Servo, but I'd like to implement it differently. Maybe instead of implementing a specific method, we could just add a `canGoBack` / `canGoForward` property to the `mozbrowserlocationchange` event. @Gozala what do you think?

paulrouget:
Browser API change: https://github.com/servo/servo/pull/10100

gordonbrander:
:+1: My strong inclination is to only show it if you can go back, but I think "greying it out" or making it very transparent is a reasonable approach as well.

paulrouget:
> My strong inclination is to only show it if you can go back

I'm ok with that.

paulrouget:
I've implemented canGoBack / canGoForward in Servo. I've updated the MDN documentation: https://developer.mozilla.org/en-US/docs/Web/Events/mozbrowserlocationchange

It hasn't landed yet. But it's been r+'ed.

paulrouget:
Landed: https://github.com/servo/servo/commit/6577409b95bc782c0493cc5186fa4ed0d31d680d

paulrouget:
New Servo build include this change.

Can someone look at this?

PR #967, opened 2016-03-28
Hide back button if you can't go back

gordonbrander:
@Gozala can you review this? I didn't remove the Gecko `canGoBackChanged` code path, but augmented it.

@paulrouget This works in Gecko, but I see no change in Servo. For Servo we should have `canGoBack` and `canGoForward` in `event.detail` (per https://github.com/servo/servo/commit/6577409b95bc782c0493cc5186fa4ed0d31d680d), but the latest servo-binary build doesn't seem to have this feature yet?

Fixes https://github.com/browserhtml/browserhtml/issues/941.

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -125,7 +125,7 @@ export const LoadEnd =
   ({type: 'LoadEnd', time});
 
 export const LocationChanged =
-  (uri/*:URI*/, canGoBack/*:boolean*/, canGoForward/*:boolean*/, time/*:Time*/)/*:Action*/ =>
+  (uri/*:URI*/, canGoBack/*:?boolean*/, canGoForward/*:?boolean*/, time/*:Time*/)/*:Action*/ =>
   ({type: 'LocationChanged', uri, canGoBack, canGoForward, time});
 
 export const ContextMenu =
@@ -934,7 +934,7 @@ const decodeLocationChange = ({detail}) =>
   // In Gecko, detail is a string (the uri).
   // In Servo, detail is an object {uri,canGoBack,canGoForward}
   ( typeof detail === 'string'
-  ? LocationChanged(detail, false, false, performance.now())
+  ? LocationChanged(detail, null, null, performance.now())
   : LocationChanged
     ( detail.uri
     , detail.canGoBack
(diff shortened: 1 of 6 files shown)


src/browser/web-view.js
```diff
@@ -919,18 +929,19 @@ const decodeError = compose(ReportError, decodeDetail);
...
   // In Servo, detail is an object {uri,canGoBack,canGoForward}
-  if (typeof detail == 'string') {
-    uri = detail;
-  } else {
-    uri = detail.uri;
-  }
-  return LocationChanged(uri, performance.now());
-}
+  ( typeof detail === 'string'
+  ? LocationChanged(detail, false, false, performance.now())
```
Gozala: This seems little odd. Usually when location changes that means that you can actually go back because you navigated from place `A` to place `B`. I would suggest to instead make types of the `canGoBack` and `canGoForward` fields `?boolean` and set those to `null` in gecko case. So we only modify `navigation.canGoBack` / `navigation.canGoForward` when action has boolean values for corresponding fields.

Gozala:
You forgot to update type signatures which is why travis build is breaking.

gordonbrander:
@Gozala fixed type signatures.

src/browser/web-view.js
gordonbrander: Ok, makes sense.

src/browser/web-view/navigation.js
```diff
@@ -155,16 +155,29 @@ export const update =
...
+          , canGoForward: action.canGoForward
+          }
+        )
+      : merge
+        ( model
+        , { currentURI: action.uri
+          }
+        )
       )
+      , Effects.batch
```
Gozala: Optional: We should not schedule `canGoBack` / `canGoForward` tasks if we already got values for them with an action.

paulrouget:
> but the latest servo-binary build doesn't seem to have this feature yet?

It does now.

paulrouget:
New build still has an issue with `goBack()`. Respinning a build.

paulrouget:
> New build still has an issue with goBack(). Respinning a build.

Done.

paulrouget:
(PR appears to work with Servo)

gordonbrander:
Ok, addressed @Gozala's last point. @paulrouget I'm still not seeing this work in latest servo-binary, but I'm going to merge as it works in Gecko and causes no harm in Servo. Let me know if I need to follow up.

Issue #968, opened 2016-03-28
show-tabs and new tab button are cut off when sidebar pinned


gordonbrander:
This is a regression caused by https://github.com/browserhtml/browserhtml/pull/943. We should either be using percentages here, or changing the hardcoded width when sidebar is pinned.

PR #969, opened 2016-03-28
Size webviews to width/height of container

gordonbrander:
Fixes https://github.com/browserhtml/browserhtml/issues/968.

@paulrouget r?


paulrouget:
r+

PR #971, opened 2016-03-29
Implement basic force-push to show tabs

gordonbrander:
This is a simple implementation of "force touch to show tabs".

This PR does **not** attempt to use `servoMouseForceChanged` to follow force 1:1. Instead, it uses a fixed-duration animation. This is because animation code for show-tabs transition assumes hard-coded duration, and will require some non-trivial refactoring. I suggest that we merge this PR first, then do a follow-up PR.

This PR fixes some problems with https://github.com/browserhtml/browserhtml/pull/946.

Fixes https://github.com/browserhtml/browserhtml/issues/872.

cc @paulrouget.

### DIFF
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -87,11 +87,11 @@ export const Focus/*:Action*/ =
   { type: 'Focus'
   };
 
-export const PushDown =
+export const PushDown/*:Action*/ =
   { type: 'PushDown'
   };
 
-export const PushedDown =
+export const PushedDown/*:Action*/ =
   { type: 'PushedDown'
   };
 
--- a/src/browser/web-view.js
+++ b/src/browser/web-view.js
@@ -87,6 +87,14 @@ export const Focus/*:Action*/ =
   { type: 'Focus'
   };
 
+export const PushDown =
+  { type: 'PushDown'
+  };
+
+export const PushedDown =
+  { type: 'PushedDown'
+  };
+
 export const Load =
   (uri/*:URI*/)/*:Action*/ =>
   ( { type: 'Load'
@@ -527,6 +535,11 @@ export const update =
   : action.type === "Close"
   ? close(model)
 
+  // Force push actions.
+  // We forward these up to WebViews.
+  : action.type === "PushDown"
+  ? [ model, Effects.receive(PushedDown) ]
+
   // Animation
   : action.type === "SelectAnimation"
   ? updateSelectAnimation(model, action.action)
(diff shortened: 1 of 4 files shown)


paulrouget:
As for the behavior, it works for me.

@Gozala can you see if Gordon's changes make sense?

Once this lands, I'll file an issue to have the intermediate animation.

src/browser/web-view.js
```diff
@@ -751,6 +764,10 @@ const viewFrame = (model, address) =>
     onBlur: on(address, always(BlurShell)),
     onFocus: on(address, always(FocusShell)),
     // onMozbrowserAsyncScroll: on(address, decodeAsyncScroll),
+    // @TODO send force changed actions and update webviews animation code
+    // to respond live 1:1 with push.
+    // onServoMouseForceChanged: on(address, decodeForceChanged),
+    onServoMouseForceDown: on(address, decodeForceDown),
```
Gozala: Is there reason to not put this into `html.div` returned by `view` function instead ? It's just we already have too event many handlers on iframe itself & I'd prefer not add more unless they are necessary of course.

Gozala: In fact why not just do that in the `WebViews.view` from what I can tell webView just propagates to webViews, moving event listener up there would remove the need to do that.

Gozala:
I suggested some optional changes, otherwise looks good to me r+

src/browser/web-view.js
gordonbrander: I assumed that to get correct event capture from moziframe, we need to bind event handler to moziframe. Otherwise the page can capture events before they bubble. I believe the way it works is that iframe gets first dibs on the event before page can cancel, then it bubbles up. Please correct me if I'm wrong here.

Gozala: > I assumed that to get correct event capture from moziframe, we need to bind event handler to moziframe. Otherwise the page can capture events before they bubble. I believe the way it works is that iframe gets first dibs on the event before page can cancel, then it bubbles up. Please correct me if I'm wrong here.

I don't believe page is even going to get that event honestly unfortunately whole event bubbling system is non existent across processes (content / chrome) at least not in gecko.

Either way you could always use capturing phase (third argument `true` or in our case `onServoMouseForceDownCapture`) in which case first will be triggered on container and only after will it reach the actual target. In fact if moziframes were regular iframes, you'd still had to use capture phase to capture events before iframe's inner document would get it.

paulrouget:
Forcetouch are not forwarded to frames.

paulrouget:
@Gozala does this need any tweaks before landing? If so, can you take over?

Gozala:
Let's land I can tweak things afterwards.

PR #975, opened 2016-03-31
add ::selection colors

paulrouget:
Servo just added support for the `::selection` pseudo element, but doesn't include default value. Let's add this to our stylesheet.

### DIFF
--- a/css/theme.css
+++ b/css/theme.css
@@ -19,6 +19,11 @@
   outline: 0;
 }
 
+.input-field::selection {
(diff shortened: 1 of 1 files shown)


paulrouget:
I could have omitted `.input-field` and just use `::selection`, but that would increase memory consumption according to https://github.com/servo/servo/pull/10252#issuecomment-203167917

tschneidereit:
r=me

PR #977, opened 2016-03-31
Fix endless RAF loop caused by web-view/progress.js

### LINKED ISSUE #976: When 2 tabs are open, CPU around 45% (infinite rAF loop)
With only one tab, CPU is very low, but as soon as a second tab opens, CPU goes crazy.

I'll investigate.

Gozala:
Progress.update was checking if new progress value was < 1 and increment progress or stop requesting frames and put progress to idle. Somehow on servo we still were getting extra update initiated by RAF in which case progress would be 0 we end up with restarted loop but this time it would never end since status remained idle and in that case we compute progress as 0.

This change makes sure that ticks on idle progress bar are ignored & there for no more new frames are requested.

Fixes #976

### DIFF
--- a/src/browser/web-view/progress.js
+++ b/src/browser/web-view/progress.js
@@ -12,10 +12,12 @@ import type {LoadProgress, Time, Model, Action} from "./progress"
 import {Effects, html} from 'reflex';
 import {ease, easeOutQuart, float} from 'eased';
 import {StyleSheet, Style} from '../../common/style';
-import {merge} from '../../common/prelude';
+import {merge, always} from '../../common/prelude';
 import * as Unknown from '../../common/unknown';
 
 const second = 1000;
+
+const NoOp = always({ type: "NoOp" })
 
 export const Start =
   (time/*:Time*/)/*:Action*/ =>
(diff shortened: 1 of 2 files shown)


paulrouget:
Can you add a short comment linking to this [issue](https://github.com/servo/servo/issues/10322) - then, once we managed to fix Servo's behavior, i'll make sure to remove this workaround.

paulrouget:
Otherwise, feel free to merge.

Gozala:
> Can you add a short comment linking to this issue - then, once we managed to fix Servo's behavior, i'll make sure to remove this workaround.

I don't think this change is a workaround though. I think logic was wrong. It's just happened to work because of the order in which events were occurring.

Gozala:
@paulrouget I can add `console.warn` if following happens: https://github.com/browserhtml/browserhtml/pull/977/files#diff-d05f0b481d8e6d62b7b684a17bc34e3aR167
which technically should not.

paulrouget:
> @paulrouget I can add console.warn if following happens: https://github.com/browserhtml/browserhtml/pull/977/files#diff-d05f0b481d8e6d62b7b684a17bc34e3aR167
which technically should not.

Yes please.

PR #978, opened 2016-03-31
Implement record / replay tool for browser.html

Gozala:
`cmd-alt-3` will dump a command into terminal invoking which will write a state snapshot file into disk & start a browser.html with it in a state snapshot was taken.

For example you can run following into your terminal (presumes PWD is project root):

```
echo '{"version":"0.1.5","mode":"create-web-view","shell":{"isFocused":true,"isMinimized":false,"isMaximized":false,"controls":{"close":{"isDisabled":false,"isFocused":false,"isActive":false,"isPointerOver":false,"text":""},"minimize":{"isDisabled":false,"isFocused":false,"isActive":false,"isPointerOver":false,"text":""},"toggle":{"isDisabled":false,"isFocused":false,"isActive":false,"isPointerOver":false,"isChecked":false,"text":""}}},"input":{"value":"amazon","isFocused":true,"isVisible":true,"selection":{"start":2,"end":6,"direction":"forward"}},"assistant":{"isOpen":true,"isExpanded":true,"query":"am","search":{"query":"am","size":5,"queryID":4,"limit":5,"selected":0,"matches":{"https://duckduckgo.com/?q=amazon":{"title":"amazon","uri":"https://duckduckgo.com/?q=amazon"},"https://duckduckgo.com/?q=american%20airlines":{"title":"american airlines","uri":"https://duckduckgo.com/?q=american%20airlines"},"https://duckduckgo.com/?q=amazon%20prime":{"title":"amazon prime","uri":"https://duckduckgo.com/?q=amazon%20prime"},"https://duckduckgo.com/?q=american%20express":{"title":"american express","uri":"https://duckduckgo.com/?q=american%20express"},"https://duckduckgo.com/?q=amtrak":{"title":"amtrak","uri":"https://duckduckgo.com/?q=amtrak"},"https://duckduckgo.com/?q=american%20eagle":{"title":"american eagle","uri":"https://duckduckgo.com/?q=american%20eagle"},"https://duckduckgo.com/?q=amazon%20books":{"title":"amazon books","uri":"https://duckduckgo.com/?q=amazon%20books"},"https://duckduckgo.com/?q=american%20idol":{"title":"american idol","uri":"https://duckduckgo.com/?q=american%20idol"},"https://duckduckgo.com/?q=amy%20adams":{"title":"amy adams","uri":"https://duckduckgo.com/?q=amy%20adams"},"https://duckduckgo.com/?q=AMC":{"title":"AMC","uri":"https://duckduckgo.com/?q=AMC"}},"items":["https://duckduckgo.com/?q=amazon","https://duckduckgo.com/?q=american%20airlines","https://duckduckgo.com/?q=amazon%20prime","https://duckduckgo.com/?q=american%20express","https://duckduckgo.com/?q=amtrak","https://duckduckgo.com/?q=american%20eagle","https://duckduckgo.com/?q=amazon%20books","https://duckduckgo.com/?q=american%20idol","https://duckduckgo.com/?q=amy%20adams","https://duckduckgo.com/?q=AMC"]},"history":{"query":"am","size":0,"queryID":4,"limit":5,"selected":-1,"matches":{},"items":[]},"selected":-1},"webViews":{"nextID":0,"selector":null,"order":[],"entries":{},"display":{"depth":0},"foldAnimation":null,"isFolded":true},"sidebar":{"isAttached":false,"isOpen":false,"animation":null,"display":{"x":550,"shadow":0.5,"spacing":16,"toolbarOpacity":1,"titleOpacity":1,"tabWidth":288},"toolbar":{"pin":{"isDisabled":false,"isFocused":false,"isActive":false,"isPointerOver":false,"isChecked":false,"text":""},"close":{"isDisabled":false,"isFocused":false,"isActive":false,"isPointerOver":false,"text":""}}},"overlay":{"isCapturing":false,"isVisible":false,"animation":null,"display":{"opacity":0}},"devtools":{"isActive":false,"settings":null},"resizeAnimation":null,"display":{"rightOffset":0},"isExpanded":true}' > ./dist/snapshot-1af7gd9fj.json && servo -w -b --pref dom.mozbrowser.enabled http://localhost:6060\?replay\=snapshot-1af7gd9fj.json
```

This will load browser.html in a state I made a snapshot. Alternatively you could generate similar command by executing `cmd-alt-3`.

### DIFF
--- a/src/browser/index.js
+++ b/src/browser/index.js
@@ -32,6 +32,7 @@ const logger = (update) => (model, action) => {
 }
 
 const isReload = window.application != null;
+const isReplay = Runtime.env.replay != null;
 
 // If hotswap change address so it points to a new mailbox &r
 // re-render.
@@ -42,17 +43,40 @@ if (isReload) {
 document.body.classList.toggle('use-native-titlebar',
                                Runtime.useNativeTitlebar());
 
+const replay =
+  () => {
+    const request = new XMLHttpRequest();
+    // We use sync XHR as we have nothing else to render otherwise.
+    request.open
+    ( 'GET'
+    , String(Runtime.env.replay)
+    , false
+    );
+    request.overrideMimeType('text/plain');
+    request.send();
+
+    if (request.status === 200 || request.status === 0) {
+      return [JSON.parse(request.responseText), Effects.none]
+    }
+    else {
+      return UI.init()
+    }
+  }
+
 const restore =
   () =>
   [ window.application.model.value
   , Effects.none
   ]
 
+
 const application = start({
   flags: void(0),
   init:
   ( isReload
   ? restore
+  : isReplay
+  ? replay
   : UI.init
   ),
   update:
--- a/src/browser/index.js
+++ b/src/browser/index.js
@@ -7,6 +7,7 @@
 import "babel-polyfill";
 import {start, Effects} from "reflex";
 import * as UI from "./perspective-ui";
+import {playback} from '../devtools/playback';
 import {version} from "../../package.json";
 import * as Config from "../../browserhtml.json";
 import * as Runtime from "../common/runtime";
@@ -43,49 +44,36 @@ if (isReload) {
 document.body.classList.toggle('use-native-titlebar',
                                Runtime.useNativeTitlebar());
 
-const replay =
-  () => {
-    const request = new XMLHttpRequest();
-    // We use sync XHR as we have nothing else to render otherwise.
-    request.open
-    ( 'GET'
-    , String(Runtime.env.replay)
-    , false
-    );
-    request.overrideMimeType('text/plain');
-    request.send();
-
-    if (request.status === 200 || request.status === 0) {
-      return [JSON.parse(request.responseText), Effects.none]
-    }
-    else {
-      return UI.init()
-    }
-  }
-
 const restore =
   () =>
   [ window.application.model.value
   , Effects.none
   ]
 
+const middleware =
+  ( isReplay
+  ? playback
+  : x => x
+  )
 
-const application = start({
-  flags: void(0),
-  init:
-  ( isReload
-  ? restore
-  : isReplay
-  ? replay
-  : UI.init
-  ),
-  update:
-  ( Config.logging
-  ? logger(UI.update)
-  : UI.update
-  ),
-  view: UI.view
-});
+
+const application = start
+  ( middleware
+    ( { flags: void(0)
+      , init:
+        ( isReload
+        ? restore
+        : UI.init
+        )
+      , update:
+        ( Config.logging
+        ? logger(UI.update)
+        : UI.update
+        )
+      , view: UI.view
+      }
+    )
+  );
 
 
 
--- a/src/common/devtools.js
+++ b/src/common/devtools.js
@@ -60,6 +60,10 @@ export const Restart/*:Action*/ =
   { type: "Restart"
   };
 
+export const Snapshot/*:Action*/ =
+  { type: "Snapshot"
+  };
+
 const Report = result =>
   ( { type: "Report"
     , result: result
@@ -134,6 +138,7 @@ const changeSetting = (model, {name, value}) =>
     .map(SettingsAction)
   ];
 
+
 const report =
   (model/*:Model*/, result/*:Result<any, any>*/)/*:[Model, Effects<Action>]*/ =>
   [ model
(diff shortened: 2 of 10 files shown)


paulrouget:
oh man, this is awesome. It's gonna be really really useful!

src/browser/index.js
```diff
@@ -42,17 +43,40 @@ if (isReload) {
 document.body.classList.toggle('use-native-titlebar',
                                Runtime.useNativeTitlebar());
 
+const replay =
+  () => {
+    const request = new XMLHttpRequest();
+    // We use sync XHR as we have nothing else to render otherwise.
```
paulrouget: That works with Servo, but in Gecko, this still spins the event loop (yeah, gecko does that with sync xhr and alerts).  Not sure how that affects the rest of our code, but this doesn't block the event loop.

paulrouget: What's the downside of using an async xhr?

src/browser/index.js
```diff
@@ -42,17 +43,40 @@ if (isReload) {
...
+    request.open
+    ( 'GET'
+    , String(Runtime.env.replay)
+    , false
+    );
+    request.overrideMimeType('text/plain');
+    request.send();
+
+    if (request.status === 200 || request.status === 0) {
+      return [JSON.parse(request.responseText), Effects.none]
```
paulrouget: Instead, can you use   `xhr.responseType = 'json'` and then `return xhr.response`?

src/browser/index.js
Gozala: > What's the downside of using an async xhr?

I would just have to wrap most of the what happens in this file in a function and call it on response. Also as I recall at least in gecko we had to draw something right away.

It's also kind of just debug tool and there is not much we can do until we get response so...

I can change it to use async if you want.

paulrouget: I'd prefer. It's just about getting the same behavior with gecko and servo. But if it's too much work, just forget about it.

src/common/runtime.js
```diff
@@ -175,3 +177,30 @@ const platform = OS.platform()
 export const useNativeTitlebar =
   ()/*:boolean*/ =>
   platform != "darwin";
+
+const Env =
+  () => {
+    const env = Object.create(null);
+    const url = URL.parse(window.location.href);
+    const params = url.searchParams;
+    // @Hack: On servo searchParams don't show up.
```
paulrouget: Add a link to https://github.com/servo/servo/issues/10335

src/browser/index.js
Gozala: > I'd prefer. It's just about getting the same behavior with gecko and servo. But if it's too much work, just forget about it.

This actually does not work on gecko, because we can't pass query arguments there, we use url to manifest instead :) Honestly I want to change an interaction quite a bit which is why I did not wanted to bother with this too much.

src/common/devtools.js
```diff
@@ -145,6 +150,25 @@ const report =
...
 
+const snapshot =
+  (model) => {
+    const id = Date.now().toString(32)
+    const message =
+      [ `\n\n`
+        , `echo '${JSON.stringify(window.application.model.value)}' > `
+        , `./dist/snapshot-${id}.json`
+        , ` && `
+        , `servo -w -b --pref dom.mozbrowser.enabled http://localhost:6060\?replay\=snapshot-${id}.json`
```
paulrouget: Bash won't like the question mark. Add `"` around the url or double escape the question mark.

paulrouget: Up to you, but I think it's not useful  to dump "echo" and "servo". What about just dump the JSON?

Gozala: > Bash won't like the question mark. Add " around the url or double escape the question mark.

quasi strings actually preserve `\` character, example I've added into description is actually what I copied from terminal.

Gozala: > Up to you, but I think it's not useful to dump "echo" and "servo". What about just dump the JSON?

No preference, only reason I chose this was because first time I stored generated json file into root instead of `./dist` and it took me a while to figure why it did not worked, I though that would eliminate that confusion. I can switch back to just dumping `json` if you like.

In fact in longer term I wanted to just add a hook to the server that already serves files that would allow my to save dumps as json on a disk.

src/common/devtools.js
```diff
@@ -145,6 +150,25 @@ const report =
     )
   ];
 
+const snapshot =
+  (model) => {
+    const id = Date.now().toString(32)
+    const message =
+      [ `\n\n`
+        , `echo '${JSON.stringify(window.application.model.value)}' > `
+        , `./dist/snapshot-${id}.json`
```
paulrouget: What about dumping this file in `/tmp/`?

Personnaly, I don't want to pollute `dist`. I will use this features like this:

`servo -w -b --pref dom.mozbrowser.enabled "file:///Users/paul/git/browserhtml/dist/index.html?replay=/tmp/snapshot-1af86t6v7.json"`

Yeah, like I said, I would not add `echo` and `servo` in the output.

paulrouget:
This is an awesome feature. This is gonna help so much! Ideally, we would also get a feature to upload the state online on a gist for easy sharing. But that could be done in a followup :)

paulrouget:
feel to merge.

src/common/devtools.js
Gozala: > servo -w -b --pref dom.mozbrowser.enabled "file:///Users/paul/git/browserhtml/dist/index.html?replay=/tmp/snapshot-1af86t6v7.json"

Hmm, I would expect XHR to be disallowed in that case.

Anyway I'll just dump JSON ;)

src/browser/index.js
Gozala: > Instead, can you use xhr.responseType = 'json' and then return xhr.response?

This causes:

> InvalidAccessError: The object does not support the operation or argument.

Gozala: Reported as servo/servo#10346

PR #980, opened 2016-04-05
Make dev tools plugable via query args

Gozala:
1. `&replay=snapshot.json` Loads snapshot of the state.
2. `&record` - Enables snapshot taking tool. If enabled `meta-alt-3` dumps snapshot to the output & `meta-alt-4` uploads snapshot as gist and print url to it.
3. `&log` or `&log=json` enables logging of app actions.


paulrouget:
Make sure to document that somewhere.

r=me

PR #984, opened 2016-04-07
Make browser.html usable on electron.

Gozala:
With this changes you can run browser.html on top of Electron. To do so you need, to:

1. Install electron on your machine. You can do it by running `npm install -g electron-prebuilt`
2. Run browser.html on top of Electron from project root via command `electron .`

### DIFF
--- a/src/browser/web-view/electron-frame.js
+++ b/src/browser/web-view/electron-frame.js
@@ -1,0 +1,148 @@
+/* @flow */
+
+import {Effects, node, html, forward} from 'reflex';
+import * as URL from '../../common/url-helper';
+import * as Driver from 'driver';
+import * as Style from '../../common/style';
+import {on} from 'driver';
+import {always} from '../../common/prelude';
+
+
+/*::
+import type {Address, DOM} from "reflex"
+import type {ID, URI, Time, Display, Options, Model, Action} from "../web-view"
+import {performance} from "../../common/performance"
+*/
+
+const Blur = always({ type: "Blur" });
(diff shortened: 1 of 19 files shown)


src/browser/web-view/electron-frame.js
```diff
@@ -0,0 +1,148 @@
...
+  ( event ) =>
+  ( { type: "LoadFail"
+    , time: performance.now()
+    , reason: event.errorDescription
+    , code: event.errorCode
+    }
+  );
+
+// @TODO: We are missing "Connected" event and I think 'did-get-response-details'
+// maybe it in electron.
```
paulrouget: It is.

src/common/keyboard.js
```diff
@@ -13,6 +13,94 @@ import * as OS from './os';
...
+      const key =
+        ( charCode === 13
+        ? keyTable[charCode]
+        : String.fromCharCode(charCode)
+        );
+
+      return key
+    }
+    else if (this.type === 'keyup' || this.type === 'keydown') {
+      return keyTable[this.keyCode]
```
paulrouget: We should never use `keyCode`. What is that for?

Gozala: > We should never use keyCode. What is that for?

Webkit does not yet implement `KeyboardEvent.prototype.key` & this is a polyfill for Electron to handle keybindings, only defined if `KeyboardEvent.prototype.key` is not defined.

paulrouget:
r+

PR #986, opened 2016-04-07
Avoid errors on incomplete URLSearchParams implementations

### LINKED ISSUE #985: searchParams fails with Servo (and maybe webkit?)
Servo now includes `searchParams`, but it fails with: `searchParams[Symbol.iterator] is not a function`. Maybe because it's an older spidermonkey. Not sure.

Anyway, this also appears to fail with Chrome:

```js
var searchParams = new URL("http:x.com/?x=y").searchParams;
for (var p of searchParams) {
  console.log(p);
}
```

Gozala:
- WebKit does not even have `window.URLSearchParams` so attempt to construct was throwing an error.
- Servo does no implement iteration protocol for `URLSearchParams[Symbol.iterator]` so attempt to iterate was throwing.

Fixes #985

### DIFF
--- a/src/common/runtime.js
+++ b/src/common/runtime.js
@@ -189,7 +189,7 @@ const Env =
   () => {
     const url = URL.parse(window.location.href);
     const params = url.searchParams;
-    try {
+    if (params != null && Symbol.iterator in params) {
       const env = Object.create(null);
       for (let [key, value] of params) {
         if (env[key] == null) {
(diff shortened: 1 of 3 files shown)


src/common/runtime.js
```diff
@@ -187,14 +187,10 @@ export const useNativeTitlebar =
...
   () => {
-    const env = Object.create(null);
     const url = URL.parse(window.location.href);
     const params = url.searchParams;
-    // @Hack: On servo searchParams don't show up (See  servo/servo#10335)
-    if (params == null) {
-      return QueryString.parse(url.search.substr(1));
-    }
-    else {
+    try {
```
tschneidereit: The only way this would throw is because of the missing `Symbol.iterator`, right? If so, I'd prefer to test for that instead of the `try` block:
```js
if (params[Symbol.iterator]) {
  ..
} else {
  return QueryString.parse(url.search.substr(1));
}
```

tschneidereit:
r=me with the comment addressed.

PR #987, opened 2016-04-07
Add some UI to record & replay devtools.

Gozala:
1. Record tool flashes UI when you print / publish snapshot and displayed message telling you that.
2. Replay tool displays message informing that it is: loading loading snapshot or failed to load snapshot.
3. Utilizes thunks to reduce overhead of tools when not in active use.

### DIFF
--- a/src/devtools/record.js
+++ b/src/devtools/record.js
@@ -1,10 +1,11 @@
 /* @flow */
 
-import {Effects, Task, html, forward} from "reflex"
+import {Effects, Task, html, thunk, forward} from "reflex"
 import {merge, always} from "../common/prelude"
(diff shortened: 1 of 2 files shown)


paulrouget:
Sorry, took me some time to realise that `&record` was necessary (we should really document all these options).

Looks good to me. Maybe you can resize the font a bit (see http://i.imgur.com/sNJWr8s.png) and replace "..." with "…".

r+

Gozala:
> Looks good to me. Maybe you can resize the font a bit (see http://i.imgur.com/sNJWr8s.png) and replace "..." with "…".
> 
> r+

I'll do it in a followup.

PR #988, opened 2016-04-07
Improve keyboard bindings polyfill

Gozala:
Lot of our keybindings were triggered on keyup / keydown and polyfill was only handling keypress. This fixes bunch of keybindings for Electron.


paulrouget:
r+

PR #989, opened 2016-04-07
Update file layout

Gozala:
This change moves files around and updates import paths. Main reason is to use consistent “fractal layout” for source files.


Gozala:
I will land this change without review since it only changes file layout & I would like to base my followup changes on top, without having to resolve conflicts or be blocked for a day.

PR #991, opened 2016-04-08
Fix issue with missing event handlers

Gozala:
In order to set listeners to non-standard DOM events `on` function must be used instead of plain function. This caused us to miss out some events.

@paulrouget this is also why force touch was not working.


paulrouget:
r+

PR #992, opened 2016-04-08
Fix connection between window & it’s opener

Gozala:
To make `window.open(uri).close()` we relied on bunch of fragile hacks that broke at some point. This patch removes some of the hacks and instead adds several tasks to make avoid platform limitations:

forceRender - Forces renderer to update DOM in the same tick (which is needs to be a case for mozbrowseropenwindow)

replaceElement(selector, element) - Replaces element matching passed query selector with passed element. It also copies over attributes, properties & event listeners from element being replaced.

forceReplace(selector, element) - Forces render in the same tick & replaced element

Patch also removes some of the obsolete code that has being long replaced by other tasks.

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -526,6 +526,7 @@ const openWebView = model =>
       , name: ''
       , features: ''
       , ref: null
+      , guestInstanceId: null
       }
     )
   );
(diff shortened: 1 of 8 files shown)


tschneidereit:
I just reviewed this in #993, so r=me with same comments.

PR #993, opened 2016-04-08
Add type annotations for driver module

Gozala:
Add type annotations for driver module.

With this change flow will be able to ensure type check uses of `on(address, eventToAction)` or more specifically it will ensure that `eventToAction` returns `action` that is compatible with a component receiving it.

    This uncovered few existing bugs that this patch also fixes.

    Special `driver` module had to get renamed to `@driver`, because `.flowconfig` mapping was incorrectly mapping stuff in node_modules that contained `driver`.

@paulrouget Please note that this also contained changes from #992 as I did these wile working on it, but afterwards decided to split into two different pulls. In other words #992 needs to land first.


src/browser/web-view/electron-frame.js
```diff
@@ -34,6 +34,16 @@ export const view =
     , 'data-current-uri': model.navigation.currentURI
     , 'data-name': model.name
     , 'data-features': model.features
+    // Stock electron does not actually connect window with it's opener in any
```
tschneidereit: nit: "a window", here and in the second-to-next sentence.

src/driver/virtual-dom/index.js
```diff
@@ -334,81 +202,96 @@ export const selection = metaProperty((node, next, previous) => {
...
-    if (next == null || next.type !== this.type) {
-      node.removeEventListener(LOCATION_CHANGE,
-                               this.constructor.handleEvent)
-    }
-  }
-}
 
-export const onCanGoBackChange = (address, decode) => {
-  const id = `onCanGoBackChange:${hash(decode)}`
+// @Hack: Below three functions have being copied from:
```
tschneidereit: nit: s/Below/The following/

tschneidereit:
r=me with nits addressed. Thanks, the lack of types for the driver was unfortunate.

PR #995, opened 2016-04-12
Create an option for standalone progressbar

Gozala:
If browser.html is passed query argument `progressbar=standalone` it will not animate progress with regular update loop, instead it will spawn up separate animation frame loop & update element.style directly without causing update or render loops.


paulrouget:
From *682ms*, with the standalone code we go down to *493ms* and with a rAF workaround down to *391ms*. 

This is getting better :)

Gozala:
> From 682ms, with the standalone code we go down to 493ms and with a rAF workaround down to 391ms.

This is little confusing in what time frame are those numbers ? Just looking at the numbers I get an impression it's not so great.

BTW I also noticed that there are bunch of other events going on during page loads which could be a reason it's still not very fast.

paulrouget:
XXXXms per second on a 20s time of sampling while page is loading.
And yes, it's not great.

Electron (4 processes). Same order of magnitude.

Before your patch:
```
Electron 314.45
Electron  93.25        
Electron 77.40
Electron  6.10        
```

After:
```
Electron 194.67
Electron 95.09
Electron 78.91
Electron 6.79
```

paulrouget:
> BTW I also noticed that there are bunch of other events going on during page loads which could be a reason it's still not very fast.

Like?

paulrouget:
r+

Gozala:
> Like?

`LoadStart, LocationChange, Connected, MetaChanged, OverflowChange, TitleChange, IconChange, ....`

PR #996, opened 2016-04-12
Add option to use timer based requestAnimationFrame

Gozala:
With this change if `raf=timer` query argument is passed, custom `setTimeout` based `requestAnimationFrame` implementation is used. This makes it possible to avoid overhead of `requestAnimationFrame` implementation currently present in servo. For details see: servo/servo#10050


paulrouget:
r+

PR #1003, opened 2016-04-14
Change all uses of Task.future to Task.create

### LINKED ISSUE #974: Warning: Task.future is deprecated API use Task.create instead
My terminal is flooded with this message.

meehow:
fixes #974

### DIFF
--- a/src/about/repl/repl/host/window.js
+++ b/src/about/repl/repl/host/window.js
@@ -70,7 +70,7 @@ const evalContext =
 
 export const evaluate =
   (id/*:ID*/, code/*:string*/)/*:Task<Never, EvaluationResult>*/ =>
-  Task.future(() => new Promise(resolve => {
+  Task.create(() => new Promise(resolve => {
     try {
       const out = executeWith(evalContext, () => window.eval(code));
       evalContext.out[id] = out;
--- a/src/about/repl/repl/host/window.js
+++ b/src/about/repl/repl/host/window.js
@@ -70,7 +70,7 @@ const evalContext =
 
 export const evaluate =
   (id/*:ID*/, code/*:string*/)/*:Task<Never, EvaluationResult>*/ =>
-  Task.create(() => new Promise(resolve => {
+  Task.create(resolve => {
     try {
       const out = executeWith(evalContext, () => window.eval(code));
       evalContext.out[id] = out;
@@ -80,4 +80,4 @@ export const evaluate =
       evalContext.out[id] = exception;
       resolve(error(exception));
     }
-  }))
+  })
--- a/src/browser/pallet.js
+++ b/src/browser/pallet.js
@@ -66,9 +66,7 @@ export const requestCuratedColor =
   (uri/*:URI*/)/*:Task<Never, ?Theme>*/ =>
   Task.create(() => {
     const hostname = getDomainName(uri);
-    return Promise.resolve
-      ( hostname == null
-      ? null
-      : curated[hostname]
-      );
+    return hostname == null
+      ? undefined
+      : curated[hostname];
   });
--- a/src/browser/pallet.js
+++ b/src/browser/pallet.js
@@ -64,7 +64,7 @@ export const create =
 
 export const requestCuratedColor =
   (uri/*:URI*/)/*:Task<Never, ?Theme>*/ =>
-  Task.future(() => {
+  Task.create(() => {
     const hostname = getDomainName(uri);
     return Promise.resolve
       ( hostname == null
--- a/src/browser/web-view/shell.js
+++ b/src/browser/web-view/shell.js
@@ -51,7 +51,7 @@ const ZoomChanged =
   ({type: "ZoomChanged", zoomResult: result});
 
 const setZoom = (id, level) =>
-  Task.future(() => {
+  Task.create(() => {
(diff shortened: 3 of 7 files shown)


Gozala:
Thanks a lot @meehow I'm so glad we can finally get rid of these warnings! There is two little changes I would like to ask you to make before I can merge this one in.

1. Could you please change `Task.create` to `new Task` instead ? API is the same & we'll likely kill `Task.create` to make more straight forwards by matching `Promise` API more closely. I feel so sorry for the misleading warning in the console. If you don't want to do it that's ok as well, we can defer that until `Task.create` is deprecated as well.

2. There is a subtle difference between `Task.create / new Task` and former `Task.future`. Since `Task.future` was returning a promise that implied that task would succeed only on next turn of event loop. Since your changes just call `resolve` in some instances (where we don't actually perform async IO)  task may succeed in the same turn of event loop which may introduce races. In a future, we'll probably make all tasks succeed on next tick unless explicitly requested otherwise, but until then we should really avoid succeeding tasks on the same tick, which is why I'd like to ask you to change your patch to ensure that is the case. I'm also pointing out individual instances where things need to be wrapped into promises  to avoid changes in behavior.

Thanks

src/about/repl/repl/host/window.js
```diff
@@ -70,7 +70,7 @@ const evalContext =
 
 export const evaluate =
   (id/*:ID*/, code/*:string*/)/*:Task<Never, EvaluationResult>*/ =>
-  Task.future(() => new Promise(resolve => {
+  Task.create(resolve => {
```
Gozala: We need to keep a promise wrapper here in order to keep an async resolution behavior as it was prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

src/browser/pallet.js
```diff
@@ -64,11 +64,9 @@ export const create =
 
 export const requestCuratedColor =
   (uri/*:URI*/)/*:Task<Never, ?Theme>*/ =>
-  Task.future(() => {
+  Task.create(() => {
```
Gozala: We need to keep a promise wrapper here in order to keep an async resolution behavior as it was prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

src/browser/web-view/shell.js
```diff
@@ -51,7 +51,7 @@ const ZoomChanged =
   ({type: "ZoomChanged", zoomResult: result});
 
 const setZoom = (id, level) =>
-  Task.future(() => {
+  Task.create(resolve => {
```
Gozala: We need to keep a promise wrapper here in order to keep an async resolution behavior as it was prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

src/browser/web-view/shell.js
```diff
@@ -86,7 +86,7 @@ export const resetZoom =
 
 export const setVisibility =
   (id/*:ID*/, isVisible/*:boolean*/)/*:Task<Never, Result<Error, boolean>>*/ =>
-  Task.future(() => {
+  Task.create(resolve => {
```
Gozala: We need to keep a promise wrapper here in order to keep an async resolution behavior as it was prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

src/common/runtime.js
```diff
@@ -84,7 +84,7 @@ export const never/*:Task<Never, any>*/ =
 
 export const respond = /*::<message>*/
   (message/*:message*/)/*:Task<Never, message>*/ =>
-  Task.future(() => Promise.resolve(message));
+  Task.create(resolve => resolve(message));
```
Gozala: We need to keep a promise wrapper here in order to keep an async resolution behavior as it was prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

src/common/settings.js
```diff
@@ -58,32 +58,31 @@ const merges =
...
-  Task.future(() => {
+  Task.create(resolve => {
     if (navigator.mozSettings != null) {
       const lock = navigator.mozSettings.createLock();
       const settings = names.map(name => lock.get(name));
-
-      return Promise.all(settings)
-                    .then(merges)
-                    .then(ok, error);
+      Promise.all(settings)
```
Gozala: Nit: Could you please move `.all(settings)` on the next line as well since you're changing indentation anyhow.

src/common/settings.js
```diff
@@ -58,32 +58,31 @@ const merges =
...
-        lock
+      lock
         .set(settings)
-        .then(always(ok(settings)), error);
-      return result
+        .then(always(ok(settings)), error)
+        .then(resolve);
     } else {
-      return Promise.resolve(error(NotSupported));
+      resolve(error(NotSupported));
```
Gozala: I'm afraid we need to keep the Promise.resolve here in order to preserve async resolution as prior to this change (See https://github.com/browserhtml/browserhtml/pull/1003#issuecomment-211648650 for more details)

```

Gozala:
Closing in favor of #1018

PR #1004, opened 2016-04-19
Pass CLI args as Query args and disable chrome privileges

Gozala:


paulrouget:
Can you explain what this is supposed to do?

Gozala:
> Can you explain what this is supposed to do?

I have documented available options and how to use them on servo or Electron here:
https://github.com/browserhtml/browserhtml/wiki/Browser-options

This change passes CLI arguments passed to Electron to a browser.html as query arguments so that
 `electron . --progressbar standalone` will be equivalent of `servo -w -perf dom.mozbrowser.enabled http://localhost:6060\?progressbar=standalone`.

paulrouget:
r=me

PR #1013, opened 2016-05-04
Use section elements instead of embed.

Gozala:
embed was terrible choice that also happened to not work on Electron.


paulrouget:
r+ but we could just for for `div`.

PR #1014, opened 2016-05-04
Fix for `isNotURL` function on Electron.

Gozala:
Electron does not throw on `new URL(‘foo bar’)` which caused misbehavior on Electron.

### DIFF
--- a/src/common/url-helper.js
+++ b/src/common/url-helper.js
@@ -117,8 +117,8 @@ export const isNotURL = (input/*:string*/)/*:boolean*/ => {
     str = 'http://' + str;
   }
   try {
-    new URL(str);
-    return false;
+    // Electron does not throw on `new URL('foo bar')`.
+    return new URL(str).hostname.includes('%20');
   } catch (e) {
     return true;
   }


paulrouget:
That's surprising. In Chrome, it does throw. And why are you testing `%20`?

Gozala:
> That's surprising. In Chrome, it does throw. And why are you testing %20

```js
encodeURI(' ') // => "%20"
```

In electron `new URI('foo bar'). hostname` is `"foo%20bar"`

Gozala:
Note in longer term I'd like to throw away our `url-helpers` module in favor of https://www.npmjs.com/package/url that implements same API as same named core module in node.

paulrouget:
> In electron new URI('foo bar'). hostname is "foo%20bar"

I get that. But why do you test the presence of a space? That's not enough. For example, `foo|bar` is not a URL. What about using a url regex?

Gozala:
> I get that. But why do you test the presence of a space?

Oh so, a lot of times we get completions from search engine like "foo bar" (often times I just type multiple words into location bar as well) when hitting enter on Electron that tries to load `http://foo%20bar` and fails to load. So this was mainly to fix that.

> That's not enough. For example, `foo|bar` is not a URL.

That is correct but from my testing space is only thing that `new URL` does not seem to throw exception on.

> What about using a url regex?

That function already uses three regexeps that I don't fully understand, I really don't want add one more in attempt to get it right. Instead I would like to replace this module with one that actually parses urls, but at this point I just want to fix the bug that I run into a lot.

paulrouget:
ok!

paulrouget:
r+

PR #1015, opened 2016-05-04
UI for session record

Gozala:
sessionrecord

---

This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1015)


Gozala:
Depends on #1021

Gozala:
This change includes changes from #1021 so it would make sense to review after that one lands to reduce already large enough diff.

Even though this change does not make record / replay fully functional, I still would really like to land this sooner than later as:

1. It has some additional functionality that I find helpful during development.
2. I'd like to avoid more maintenance of this patch.
3. It introduces better effect management primitives that I'd like to use in other modules.

Gozala:
## Overview (in no particular order)

- New modules are now named with different convention like `ToggleButton.js` vs `toggle-button.js` as usually imports look like `import ToggleButton from "./ToggleButton"`.
- New modules use alternative way to manage tasks. Traditionally our update functions would return `[Model, Effect<Action>]` tuple but that made composition of multiple updates more complicated & require construction / deconstruction off tuples at every component layer. Now instead effectful components have `io: IO.Model`  field and extras effects can be just queued into it. That makes update API a lot simpler, makes composition easier & supposedly more optimal since given that it's no longer to reconstruct tuples at every step.
- Few generic `UI/*.js` modules have being added for used controls.
- Focus management, Activation and friends are now handled by `UI/Contorol/*.js` modules & their users can now directly delegate to them. Also note that before focus management was done via "special hacks" provided in `@driver` module. Now these operations are no longer special and expressed via tasks.
- `Ref`s has being introduce as  a way to pass references to the actual DOM to a tasks than need them - like `Focus.js`. 
- Model's now are expressed via class instances vs plain objects. This way we avoid suboptimal generic `merge` function.

paulrouget:
This is a giant PR. Do you want to land it as it is or can you split that in different PRs? Or maybe less smaller commits.

tschneidereit:
Ok, I reviewed at least a few files. My main concern is that there's no documentation whatsoever, and large parts of this aren't self-explanatory at all. I asked for documentation in a few places below, but in general it'd be great to have at least a short explanation for each module. Without that, I'm afraid I won't be able to meaningfully sign off on this, as I won't really understand it.

<details><summary><i>Previously, paulrouget (Paul Rouget) wrote…</i></summary>
> This is a giant PR. Do you want to land it as it is or can you split that in different PRs? Or maybe less smaller commits.
</details>


---

Reviewed 9 of 29 files at r2, 1 of 3 files at r3.
Review status: 10 of 27 files reviewed at latest revision, 14 unresolved discussions, some commit checks failed.

---

*[src/devtools.js, line 10 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBGJmWlpz0zcIjqStL:-KJBGJmWlpz0zcIjqStM:-1113106858) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools.js#L10)):*
> ```JavaScript
> import * as Unknown from "./common/unknown"
> import * as Replay from "./devtools/replay"
> import * as Record from "./devtools/Record"
> ```

Either rename both `record` and `replay` to `Record` and `Replay`, or none of them.

---

*[src/devtools.js, line 47 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBJ06X-kFBB-i1CZc4:-KJBJ06Yf4L85NsQpamm:-1747624468) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools.js#L47)):*
> ```JavaScript
> 
>   Debuggee: Debuggee<input, state>;
>   debuggee: state;
> ```

We really shouldn't have two fields whose names differ just by their casing. Also, an instance of a type, in this case `Debuggee`, shouldn't have the same name as the type itself. Not sure what to call these things instead, but something else.

---

*[src/devtools.js, line 53 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBaIrMw73vThnRnH9p:-KJBaIrMw73vThnRnH9q:-261741928) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools.js#L53)):*
> ```JavaScript
>   constructor(
>     Debuggee/*:Debuggee<input, state>*/
>   , debuggee/*:state*/
> ```

This goes for the parameter, too, obviously.

---

*[src/devtools.js, line 58 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBfu4p8BNHrgMczo1Y:-KJBfu4qcDvpv7IBtqKL:-575535506) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools.js#L58)):*
> ```JavaScript
>   , record/*:?Record.Model<input, state>*/
>   , replay/*:?Replay.Model<input, state>*/
>   , log/*:?Log.Model<input, state>*/
> ```

AFAICT, most of these parameters are extracted from a single `model` 

---

*[src/devtools.js, line 190 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBdrOpmNalCCAaApqy:-KJBdrOqp7iBJYHcpHr4:848591470) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools.js#L190)):*
> ```JavaScript
>   }
> 
> const panic = /*::<message, input, state>*/
> ```

I don't think this should be called `panic`. That name sounds like calling this will abort execution, which it doesn't do.

At the very least, please add a comment explaining what this does.

---

*[src/common/IO.js, line 1 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBjPVduqv1xYDzvSIv:-KJBjPVduqv1xYDzvSIw:-332965570) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/common/IO.js#L1)):*
> ```JavaScript
> /* @flow */
> ```

This is missing a license header

---

*[src/common/IO.js, line 18 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBjTUawgFMR3zEIGW0:-KJBjTUawgFMR3zEIGW1:-613692944) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/common/IO.js#L18)):*
> ```JavaScript
> 
> 
> class IO /*::<a>*/ extends Effects /*::<a>*/ {
> ```

Short comments explaining what this and the other classes here do would be nice

---

*[src/common/IO.js, line 38 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBjgQ5dIbW8loRI6Y2:-KJBjgQ6Risay7mQrm_-:-1904704947) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/common/IO.js#L38)):*
> ```JavaScript
>         const task = queue[index]
>         Task.fork(task, address, raise)
>         index = index + 1
> ```

What's wrong with `index++`?

---

*[src/common/lens.js, line 1 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBjs7ZqS1lgBHt8I06:-KJBjs7ZqS1lgBHt8I07:-1386086684) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/common/lens.js#L1)):*
> ```JavaScript
> /* @flow */
> ```

License header. Please check this for all new files.

---

*[src/common/lens.js, line 13 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBjwZTPrN5JQOJnjtV:-KJBjwZTPrN5JQOJnjtW:-1438252668) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/common/lens.js#L13)):*
> ```JavaScript
> */
> 
> class Lens /*::<from, to>*/ {
> ```

Here, a comment is definitely required. Or a more self-explanatory name. I have no idea what "lens" means in this context.

---

*[src/devtools/Debug.js, line 1 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBkQ3VbM9NN9xA_L9q:-KJBkQ3VbM9NN9xA_L9r:2040349292) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools/Debug.js#L1)):*
> ```JavaScript
> /* @flow */
> ```

Header, plus short explanatory comment.

---

*[src/devtools/log.js, line 1 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBkW6pkxbE6v8F1L7k:-KJBkW6q_IPSYh9W1MPT:-1950587520) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools/log.js#L1)):*
> ```JavaScript
> /* @flow */
> ```

Pre-existing, but missing a license header, too.

---

*[src/devtools/log.js, line 82 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBki0xZniZd3HF6rOi:-KJBki0ySC6t64YsFhqS:1421985978) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools/log.js#L82)):*
> ```JavaScript
>   }
> 
> export const panic = /*::<message, input, state>*/
> ```

Same as above: `panic` is the wrong name for this, I think.

---

*[src/devtools/Record.js, line 1 \[r4\]](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-KJBktEsFJ7HtJ4Z1UaS:-KJBktEtLSM12MrYmKMn:83771967) ([raw file](https://github.com/browserhtml/browserhtml/blob/589d078754c96fa522fa7c9d51360ceb9df476ad/src/devtools/Record.js#L1)):*
> ```JavaScript
> /* @flow */
> ```

This really was renamed from `record.js`, right? It's a bit unfortunate that the diff is lost to the rename :(

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/browserhtml/browserhtml/1015#-:-KJBlVu0cvSo88eflUb8:1236385351)*

Gozala:
Does not render on servo properly, submitted issue servo/servo#12144

PR #1018, opened 2016-05-07
Task.future

Gozala:
I'm taking over #1003 since I got no response from contributor for a while. This patch based over #1003 but it also addresses review comments made there.

### DIFF
--- a/src/about/repl/repl/host/window.js
+++ b/src/about/repl/repl/host/window.js
@@ -70,7 +70,7 @@ const evalContext =
 
 export const evaluate =
   (id/*:ID*/, code/*:string*/)/*:Task<Never, EvaluationResult>*/ =>
-  new Task((succeed, fail) => new Promise((resolve, reject) => {
+  new Task((succeed, fail) => void(new Promise((resolve, reject) => {
     try {
       const out = executeWith(evalContext, () => window.eval(code));
       evalContext.out[id] = out;
@@ -80,4 +80,4 @@ export const evaluate =
       evalContext.out[id] = exception;
       resolve(error(exception));
     }
-  }).then(succeed, fail))
+  }).then(succeed, fail)))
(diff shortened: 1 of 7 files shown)


src/about/repl/repl/host/window.js
```diff
@@ -70,7 +70,7 @@ const evalContext =
 
 export const evaluate =
   (id/*:ID*/, code/*:string*/)/*:Task<Never, EvaluationResult>*/ =>
-  Task.future(() => new Promise(resolve => {
+  new Task((succeed, fail) => void(new Promise((resolve, reject) => {
```
tschneidereit: Why're you wrapping this in `void()`?

src/browser/assistant/history.js
```diff
@@ -61,18 +61,14 @@ const pendingRequests = Object.create(null);
 
 const abort =
   (id/*:number*/)/*:Task<Never, number>*/ =>
-  Task.future(() => new Promise(resolve => {
-
-  }));
+  new Task(succeed => void(0))
```
tschneidereit: Using `void(0)` used to be better than `undefined` back when `undefined` was configurable so could be changed to something else. That hasn't been the case for many years now, so `undefined` isn't the least bit worse. (Except for being 2 chars longer, which is why minifiers still use it.)

src/about/repl/repl/host/window.js
tschneidereit: Ah, I think I understand: otherwise a `Promise` would be returned which would cause the `Task` chain to be extended, right? I'm not sure I'm too big a fan of this pattern as it's quite subtle, but it does make sense.

tschneidereit:
Looks good. The `void(0)` could be changed, but doesn't need to, so let's merge this.

src/about/repl/repl/host/window.js
Gozala: No it's just type of Task, which is:

```js
class Task <error, value> {
  constructor(perform: (succeed:(value:value) => void, fail:(error:error) => void) => void)
  //...
}
```

So function passed to a constructor must not return non undefined, otherwise type checker will not type check.

PR #1019, opened 2016-05-09
avoid container overflow

paulrouget:
A button is 12px large. The last button is 40px from the left. 40 + 12  = 52.


tschneidereit:
Makes sense, r=me

PR #1023, opened 2016-05-09
Iframe display block

Gozala:
This change reverts d15d0ae as it was breaking layout on Electron & instead add's `display:block` to only `mozbrowser` iframes.


paulrouget:
r=me

PR #1024, opened 2016-05-09
Fix failing type check with flow 0.24.

Gozala:
Flow 0.24 started to complain about nullURL as it’s indeed not instance of `URL`. Instead of trying to workaround that `nullURL` has being moved to own module that is opted out from flow type checking.


src/common/url.js
```diff
@@ -0,0 +1,34 @@
+/* @noflow */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+export const URL = window.URL;
+export const URLSearchParams = window.URLSearchParams;
+export const blank =
```
tschneidereit: I think I preferred `nullURL` as a name. `blank` could be anything, really.

tschneidereit:
r=me with `blank` renamed to `nullURL` again. (Or with an explanation why the new name is better.)

PR #1026, opened 2016-05-10
Use 'pointer-events:auto' instead of 'all', which is not a HTML property value

paulrouget:


tschneidereit:
@paulrouget, can you add an explanation for what this fixes to the PR message? r=me with that.

paulrouget:
Done.

PR #1032, opened 2016-05-12
Fix regression in live reloading.

Gozala:
New devtools code stores module it instruments in the model, this broke live reload as application is restored from last state which still used to point to old module code that has not changed.


tschneidereit:
r=me

Issue #1027, opened 2016-05-10
Move + and < buttons to toolbar


paulrouget:

tschneidereit:
One obvious concern is that the buttons should be visible when the sidebar is either open or pinned. Can we perhaps use a filter to have them be the inverse color of the background? That way they would always be visible and we wouldn't have to worry about any transitions.

OTOH, inverting the background might not look great depending on the titlebar's content-based color.

paulrouget:
Not sure to understand. The toolbar moves with the tab when it's zooming out. So the button are always against the same background.

Gozala:
Fixed by #1047

Issue #963, opened 2016-03-25
Move input, overlay and suggestions into web-view


Gozala:
As me and @gordonbrander found out from #843 it is going to be a far more easier implement #843 & avoid dealing with too many edge cases if we move input, overlay and suggestions into web-view component itself.

To avoid too many changes with a `about:newtab` we're going to do it as separate change.

Gozala:
Fixed by #1047

Issue #843, opened 2016-01-19
Bring back "new tab"


gordonbrander:
Our first pass could be to bring back the website tiles. I love these from the standpoint of usability. It's really easy to recognize your favorite sites this way.

![artboard 187 copy 114](https://cloud.githubusercontent.com/assets/58421/12433654/2b717112-beb7-11e5-86e9-2d91e747e0f9.png)

paulrouget:
Yes!

matthewp:
This is really nice.

Gozala:
Depends on #843

Gozala:
Fixed by #1047

Issue #747, opened 2015-11-27
Switching animation should not carry on from the current display state.


Gozala:
At the moment animations compute every property that is animated from the animation progress. While this greatly simplifies state management overhead for animations it introduces a major limitation - That is:

If say you have `"show-tabs"` animation running (that animates an elements `translateX`) that supposed to have  `d1`ms duration, but after `p1`ms from the start of that animation user triggers `"hide-tabs"` animation (animates `translateX` in an opposite direction) that supposed to run for `d2`ms:

```
show-tabs: -----p1--d1
hide-tabs:      ----------d2
```

You would want to:
1. To animate `x` from the current position which will be different from the generally expected start position.
2. Adjust duration of the `"hide-tabs"` animation presumably so that it will be shorter than `d2` given that starting point of the animated element is different (distance `x` needs to travel shorter).

Initially I assumed that in such case we would generate new animation model `{start, end, time}` but offset it such that `start` progress of the new animation would match up with a progress of the previous animation. But this solution is incomplete given that progress may be translated to `x` with a different  easing in `show-tabs` / `hide-tabs` animation there for `p` progress in `show-tabs` animation may translate to `x` that is different from `y` that is computed from the same `p` progress in `hide-tabs` animation. Not to mention that different animation may animate different properties there for properties not present in the subsequent animation would have to be somehow be transitioned to normal.

In general I think there are following possible solutions:
1. We may just queue action that triggered animation switch until currently running animation is complete.  Possibly speed up the animation.
   
   **Pros:**
   1. Any combination of animations will work without explicit interference.
   
   **Cons:**
   1. UI may appear unresponsive to a user given that interaction will be queued.
   2. Several use actions may end up in a queue while first animation is still completing. In some cases it maybe `show / hide / show` and in such case we either drop first two and require logic to handle this or just go with a queue. In one case UI will appear unresponsive and in other it will appear laggy. 
2. We can store properties being animated into a model instead of computing them in the view. That way switched animation will start from the current state of the model instead of assumed starting point.
   
   **Pros:**
   1. We will switching animations right away and there for UI won't appear laggy or unresponsive.
   
   **Cons:**
   1. Each animation would have to consider all other animations it may be replaced with - meaning that they would have to take into account every property that can be animated and consider how to interpolate it to provide smooth transition (Probably applying some spring physics would make this a lot easier).
   2. Requires extra state management.

gordonbrander:
[...]

PR #1034, opened 2016-05-13
about:newtab page

Gozala:
# Work in progress / Do not merge!

Ok primary goal of this change is to bring back `about:newtab` page which was unclose-able, always first in the list tab & all the commands to open new tab would just switch to it. All the navigation from `about:newtab` would just create new tabs with a url you navigated to.

In order to implement this we drafted plan with following steps

- Move `Input`, `Overlay` and `Assistant` into `WebView` #963
- Make "about:newtab" special: unclosable, titlebarless & always first in the list #843 

As it turn out doing #963 without #843 is not ideal as it will required new logic that will be made obsolete by #843. More specifically once `Input` / `Overlay` / `Suggestions` move inside a `WebView` there will be no way to render app in a `"create-webview"` mode since there will be no top level `Input`  or `Assistant` to display. There for it seems better to utilize already existing `about:newtab` page as a special tab that will be used  to render app in `"create-webview"` mode which is pretty much what #843 is about.

## Overview of changes

- Instead of adding `Input` / `Overlay` / `Assistant` into a `WebView` new component `Navigator`  has being created that holds  `WebView`, `Input`, `Overlay` and `Assistant`. Ideas is to make it mediate these component pretty much how  `Browser` used to do it before.
- New `Navigators` component has being introduced as replacement for `WebViews` as we'll have list of `Navigator`s rather that `WebViews`. Also given that actual management of list if things is pretty generic & agnostic of `things` that list consists of generic `Deck<Card>` component had being created and `Navigators` is just wrapper that fills in details about specifics of the `Card` in this case `Navigator`.
- `NavigatorDeck` component had being added as replacement for previously hacked together [`main.root > .browser-content`](https://github.com/browserhtml/browserhtml/blob/master/src/browser.js#L841) container of `Input`, `Overlay`, `Assistant` & `WebViews` which also used to animate as you opened / closed / attached `SideBar`. Component just embeds `Deck` of `Navigators` and manages earlier mentioned  `animation` state.
- As `Input` / `Overlay` / `Suggestions` moved into ~~`WebView`~~ `Navigator` component #747 became more apparent.  There for `Animation` component has being introduced as an alternative for `Stopwatch` component that manages display animations such that starting new ones carry on from the current display state vs from the state we expect to start from.
- As Things moved out of `WebView` into `Navigator` I had to move `OpenNewTab` and `GoBack` buttons along & made little sense. Since stacking changed positions had to be assigned to buttons & it seemed to make very little sense to move them to original positions (that required extra layers) vs just moving them into a toolbar (that required no extra layers) so I went ahead with that.

## These pull request fixes

- [x] Animation should update display from the state it's in vs from the state we expect it to be #747
- [x] Move `Input`, `Overlay` and `Assistant` into `WebView` #963
- [ ] Make "about:newtab" special: unclosable, titlebarless & always first in the list #843 
- [x] Move OpenNewTab and GoBack buttons into a toolbar. #1027

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1034)


Gozala:
Close in favor of rebased version #1047

PR #1049, opened 2016-05-31
(WIP) Issue reporter

### LINKED ISSUE #1001: Catch pipeline crashes and report error
Servo will soon catch crashes in iframes and report the crash with a mozbrowser event.

See https://github.com/servo/servo/issues/10334

We need to do something about this in browserhtml.

paulrouget:
Fix #1001

TODO:
* stop loading page on crash
* we get duplicate reports
* report crash from top level pipeline too
* use a proper github repo to report issues

This is still work in progress, but I'd like to get an early review from @Gozala. Also, I don't know what is the right approach to stop the page load. Please advise.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1049)

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -25,6 +25,7 @@ import * as OS from './common/os';
 import * as Keyboard from './common/keyboard';
 import * as Stopwatch from "./common/stopwatch";
 import * as Easing from "eased";
+import * as Issue from "./common/issue-reporter";
 import {merge, always, batch, tag, tagged} from "./common/prelude";
 import {cursor} from "./common/cursor";
 import {Style, StyleSheet} from './common/style';
@@ -43,6 +44,7 @@ export const init = ()/*:[Model, Effects<Action>]*/ => {
   const [devtools, devtoolsFx] = Devtools.init({isActive: Config.devtools});
   const [input, inputFx] = Input.init(false, false, "");
   const [shell, shellFx] = Shell.init();
+  const [issue, issueFx] = Issue.init();
   const [webViews, webViewsFx] = WebViews.init();
   const [sidebar, sidebarFx] = Sidebar.init();
   const [assistant, assistantFx] = Assistant.init();
@@ -52,6 +54,7 @@ export const init = ()/*:[Model, Effects<Action>]*/ => {
     { version
     , mode: 'create-web-view'
     , shell
+    , issue
     , input
     , assistant
     , webViews
@@ -69,6 +72,7 @@ export const init = ()/*:[Model, Effects<Action>]*/ => {
     ( [ devtoolsFx.map(DevtoolsAction)
       , inputFx.map(InputAction)
       , shellFx.map(ShellAction)
+      , issueFx.map(IssueAction)
       , webViewsFx.map(WebViewsAction)
       , sidebarFx.map(SidebarAction)
       , assistantFx.map(AssistantAction)
@@ -135,6 +139,11 @@ const WebViewsAction = (action/*:WebViews.Action*/)/*:Action*/ =>
   ? ShowTabs
   : action.type === "Create"
   ? CreateWebView
+  : action.type === "Crash"
+  ? {
+    type: "ReportWebviewCrash",
+    source: action
+  }
   : action.type === "Edit"
   ? EditWebView
   : action.type === "SelectRelative"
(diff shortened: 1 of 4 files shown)


src/browser.js
```diff
@@ -168,6 +177,20 @@ const ShellAction = action =>
...
 
+const IssueAction = action =>
+  ( action.type === "DismissReport"
+   ? {
+     type: "DismissReport",
+   }
+   : action.type === "FileReport"
+   ? {
+     type: "FileReport",
+   }
```
Gozala: It seems to me that above two are redundant and you could just had:

src/browser.js
```diff
@@ -790,6 +820,11 @@ export const update =
       .map(NoOp)
     ]
 
+  : action.type === 'ReportWebviewCrash' ||
+    action.type === 'DismissReport' ||
+    action.type === 'FileReport'
```
Gozala: With the suggestion above this would become:

```js

File: src/common/issue-reporter.js [hunk 3]
```diff
@@ -0,0 +1,135 @@
...
+    return Unknown.update(model, action);
+  }
+}
+
+export const view =
+  (model/*:Model*/, address/*:Address<Action>*/)/*:DOM*/ => {
+
+  var report_available = model.reports.length > 0;
+  if (!report_available) {
+    return html.div();
```
Gozala: I'd return `""` instead.

src/common/issue-reporter.js
```diff
@@ -0,0 +1,135 @@
...
+    var report = {
+      title: action.source.description,
+      url: action.source.url,
+      backtrace: action.source.backtrace,
+    };
+    return [merge(model, {reports: [...model.reports, report]}), Effects.none]
+  } else if (action.type === "DismissReport") {
+    return [merge(model, {reports: model.reports.slice(1)}), Effects.none];
+  } else if (action.type === "FileReport") {
+    var report = model.reports.shift();
```
Gozala: Please avoid mutations and use `model.reports[0]` instead.

Gozala:
## Cancelling load / stopping progress-bar

Alright so what you want your pull does is propagates `Crash` action to the `Browser`, which then feeds it right back to `Issue` component, but what you really want (which I did not realized initially) is to handle `Crash` action in both `Issues` and a specific `WebView` component that triggered the action, so that it can be updated as no longer loading etc... This makes things slightly trickier & there are two options to go about it:

1. Carry `id` of the `WebView` that triggered the action while propagating `Crash` event so that once it loops back it could be fead back to the `WebView` that crashed.

2. Have two distinct actions one internal that just tells `update` function that crash happened and the other that gets mapped while propagating & is delegated to `Issues` once it loops back.

I personally lean towards the 2nd option as unlike 1st option you don't need to teach all the components in the middle about the fact that there are multiple `WebView`s and that each have unique ID.


In order to do what's described in 2nd option I'd do following:

1. Defined `{ type: "Failure", failure: Failure }` and `{ type: "Crash", crash: Crash }` actions.

2. `onError` I'd send `Failure` action without any of the mapping in outer components, that way it will loop back into `WebView.update` for the instance that had the failure.

3. On `Failure` action I'd do pretty much what we do on `LoadEnd` (stop progressbar etc...) & send `Crash` action back. You can send actions back from update like: `[newModel, Effects.recieve({ type: "Crash", crash: action.failure }) ]`. Your current wiring (mappings of `Crash` action) should do the rest.

paulrouget:
There's an issue with the pipeline force-crash feature where, whatever the probability is, it crashes the main pipeline, making it impossible to work on this at the moment.

See https://github.com/servo/servo/issues/11926

paulrouget:
We found a workaround for the force-crash feature.

So far the only thing I've been trying to do is to port the early PR to the new code base. But it's harder than I thought and I won't have time to get that done today.

If someone feels like taking over, here are some instructions:

To work on this issue, it's necessary to be able to crash Servo on demand. To do so, 2 things need to be done (just locally):
1) change any instance of `about:newtab` to `about:blank` (otherwise it crashes too often)
2) start Servo like this: `servo -w -b --pref dom.mozbrowser.enabled --random-pipeline-closure-probability 0.05 http://localhost:6060`

The main Servo window might crash time to time, but usually, only the first loaded tab will crash (I usually load firefox.com and I get the crash). If the first tab doesn't crash, Cmd-R a couple of time will likely crash the pipeline.

See `decodeError` in the PR to see how crash are reported. `issue-reporter.js` is probably good enough.

3 necessary changes:
1) address Irakli initial comments
2) port to new code base (the propagation of the Crash action and the update)
3) create a new repo (like https://github.com/servo/servo-panic-from-browserhtml) and point the reporter to that repo instead of https://github.com/servo/servo (we don't want the issue to be mixed-up with the normal servo issues)

asajeffrey:
The main browser.html pipeline shouldn't crash even with `--random-pipeline-closure-probability 0.05`, but any child pipelines may, including the one displaying `about:newtab`. At some point we could add URL-based whitelisting, e.g. to make sure `about:*` pages are kept alive.

Gozala:
Closing in favor of #1145

PR #1055, opened 2016-06-01
Make interact with about:newtab different from regular tabs

Gozala:
With this patch following changes are made:

1. `about:newtab` page sets a meta tag on theme changes to update titlebar so it matches the wallpaper.
2. `about:newtab` always has a input field displayed & it can't be dismissed.
3. `about:newtab` does not get's overlay when input is active.
4. It is not possible to close `about:newtab`.
5. Entering urls / search queries in `about:newtab` opens them in new tabs.

Please note this pull is based off #1065 there for it includes changes from there, in other words I'd just look at last two commits. Also all other pulls have being factored out from here and there for all changes are necessary to get expected experience.

P.S.: This is not the approach I would like to take, more specifically polluting Navigator with `isPinned` and `isInputEmbedded` and then branching logic on updates is not good approach. Instead I would like to allow different kinds of `Navigator`s and let each kind define it's own update / view logic.  But that seems like a larger change which I'd rather do as followup than block necessary fixes.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1055)

### DIFF
--- a/src/about/newtab/newtab.js
+++ b/src/about/newtab/newtab.js
@@ -136,7 +136,12 @@ export const view =
         , readWallpaper(activeWallpaper)
         )
       }
-    , [ thunk
+    , [ html.meta
+        ( { name: "theme-color"
+          , content: activeWallpaper.color
+          }
+        )
+      , thunk
         ( 'tiles'
         , Tiles.view
         , tiles
--- a/src/browser/Navigators.js
+++ b/src/browser/Navigators.js
@@ -108,6 +108,14 @@ const tagDeck =
               type: "Deck"
             , deck: action.modify
             }
+          case "Select":
+            return {
+              type: "Deck"
+            , deck:
+              { type: "Select"
+              , id: action.id
+              }
+            }
           case "Closed":
             return {
               type: "Deck"
@@ -156,6 +164,8 @@ export const init =
         }
       , assistant: true
       , overlay: true
+      , isPinned: true
+      , isInputEmbedded: true
       }
 
     const [deck, $deck] = Deck.init();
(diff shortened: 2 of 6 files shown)


paulrouget:
> about:newtab page sets a meta tag on theme changes to update titlebar so it matches the wallpaper.

Hmm, doesn't appear to work here.

Gozala:
> > about:newtab page sets a meta tag on theme changes to update titlebar so it matches the wallpaper.

> Hmm, doesn't appear to work here.

You're right, it seems that meta changes don't trigger theme updates if they happen after first pain. How about I create a separate pull for fixing that.

P.S.: It seems that there was flow in logic & theme colors were not actually ever picked up if they happened after load.

tschneidereit:
r=me, with the rest in a follow-up.

<details><summary><i>Previously, Gozala (Irakli Gozalishvili) wrote…</i></summary>
> > > about:newtab page sets a meta tag on theme changes to update titlebar so it matches the wallpaper.
> 
> 
> 
> > Hmm, doesn't appear to work here.
> 
> 
> 
> You're right, it seems that meta changes don't trigger theme updates if they happen after first pain. How about I create a separate pull for fixing that.
> 
> 
> 
> P.S.: It seems that there was flow in logic & theme colors were not actually ever picked up if they happened after load.
</details>


---

Reviewed 3 of 6 files at r1.
Review status: 3 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/browserhtml/browserhtml/1055#-:-KJH3wbPLsKvL0LJbKpH:550427090)*

Issue #1053, opened 2016-05-31
Window focus / blur misbehaves


Gozala:
If you blur the the window mac window controls remain active while they should gray out. Also when you bring focus back an newtab page input field should get selection, but that does not happen as we don't receive event to do so.

Caused by servo/servo#11526

Gozala:
Depends on #9482 I guess.

paulrouget:
Closing in favor of servo/servo#9482 (P3 at the moment)

PR #1064, opened 2016-06-01
Listen for top level events on window.

Gozala:
Electron (& WebKit in general) does not dispatch `focus` / `blur` events on the document, it does on `window` instead. We are listening on `document` instead as it used to work in servo (see #1053) but it no longer does. This patch changes code back so we listen for top level events on window again, presumably servo/servo#9482 will fix it so that events are dispatched on window instead.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1064)


paulrouget:
r=me

PR #1067, opened 2016-06-01
Remove `+` button from sidebar.

Gozala:
That button is now redundant as sidebar already has a "New Tab" tab that serves exact same purpose.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1067)


paulrouget:
That makes sense… but it feels weird.

Let's land that, but we should really revisit this UI.

PR #1072, opened 2016-06-02
Increase default window size

tschneidereit:
800x600 is very small - even the tiny Macbooks show much more. Let's increase this to something more reasonable.

r? @paulrouget

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1072)


Gozala:
@tschneidereit can you please also update resolution for electron so the behave same, it's in here:
https://github.com/browserhtml/browserhtml/blob/master/main.js#L52-L53

Gozala:
r=me

Issue #1075, opened 2016-06-02
Some tiles from about:newtab don't get clipped properly


Gozala:
In the image below you can see that Twitter and Thumblr tiles aren't rounded as expected

<img width="832" alt="screen shot 2016-06-02 at 11 21 44 am" src="https://cloud.githubusercontent.com/assets/21236/15756022/3c9b5054-28b4-11e6-8652-a5625e5b13a8.png">

Gozala:
It seems that issue is related to window size, my guess is content that overflows is no longer clipped properly. @pcwalton mentioned following in his webrender presentation:

> Typically, all display items can be batched together to minimize state changes.
Exceptions are if we run out of texture space in an atlas or have a clipping region that isn't a (rounded) rectangle.

My guess is about:newtab page falls into that exception and webrender probably has a bug that causes it to handle that exception incorrectly

Gozala:
After further investigation I have submitted following bug servo/servo#11562

Gozala:
Submitted a fix on browser.html side that avoids the issue.

Gozala:
Workaround has landed on our side and there is a bug on file for servo, there for I think we can close this.

PR #1076, opened 2016-06-02
Fix tile clipping issues on servo

Gozala:
There are clipping issue servo/servo#11562 in webrender that occurs when clipping rounder region with in the rounded region. This change removes border-radius on body that was rounder for no apparent reason & there for avoids above mentioned issues in webrender.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1076)


Gozala:
Fixes #1075

tschneidereit:
r=me

Issue #1038, opened 2016-05-17
releasing forcetouch closes the sidebar


paulrouget:

paulrouget:
This happens because we get a click on the overlay when it's released.

paulrouget:
This doesn't always happen because of a bug in Servo where the iframe get the mouse events before the overlay. See https://github.com/servo/servo/issues/11108

Gozala:
@paulrouget It is assigned to you, does that mean you're going to fix it ? Also should it be a P2 instead ?

Gozala:
This seems to be caused by servo/servo#11664 @paulrouget do you think it can be fixed in servo or should we come up with a workaround ?

Gozala:
Create a workaround #1087 that is not so bad & avoids the issue described here.

Gozala:
was fixed by #1087

PR #1087, opened 2016-06-07
Zoom in page on mousedown

Gozala:
This workaround #1038 and servo/servo#11664 that prevents us from cancelling click events when force touch is handled. With this change we react to mousedown event instead of click which fixes our issue since overlay is not displayed when zoomed in & there for we avoid causing zoom-in at the end of force touch. clicks.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1087)


paulrouget:
Thanks.

PR #1092, opened 2016-06-16
URI to URL

Coder206:
@paulrouget wanted this change as Servo is also changing over to URL https://github.com/servo/servo/issues/11450

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1092)

### DIFF
--- a/src/browser/Navigators/Navigator/WebView/MozBrowserFrame.js
+++ b/src/browser/Navigators/Navigator/WebView/MozBrowserFrame.js
@@ -185,7 +185,7 @@ const decodeLocationChange =
     , canGoForward: null
     }
   : { type: "LocationChanged"
-    , uri: detail.uri
+    , url: detail.url
     , time: performance.now()
     , canGoBack: detail.canGoBack
     , canGoForward: detail.canGoForward
--- a/src/browser/Navigators/Navigator/WebView/MozBrowserFrame.js
+++ b/src/browser/Navigators/Navigator/WebView/MozBrowserFrame.js
@@ -185,7 +185,7 @@ const decodeLocationChange =
     , canGoForward: null
     }
   : { type: "LocationChanged"
-    , uri: detail.url || detail.uri || detail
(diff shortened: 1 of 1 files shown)


Coder206:
@paulrouget I am in the final stages of fixing this Servo, maybe delay until this PR is a go.
#11789

src/browser/Navigators/Navigator/WebView/MozBrowserFrame.js
```diff
@@ -185,7 +185,7 @@ const decodeLocationChange =
     , canGoForward: null
     }
   : { type: "LocationChanged"
-    , uri: detail.uri
+    , url: detail.url
```
paulrouget: You don't want to change the decoded property name.

So more like: `uri: detail.url || detail.uri || detail` (for backward compatibility with previous version of Servo and with Gecko).

Coder206: @paulrouget Ok! That makes sense :) I noticed other lines in this file like this `uri: detail`, should I change those too?

paulrouget: Oh good catch! So, no, you don't need, because the other is for Gecko, but then, the ` || detail` that you added is not necessary.

So basically, you just need to `uri: detail.uri` to `uri: detail.url || detail.uri`.

Coder206: @paulrouget Thanks! Will do that now. As for Servo's side of the URI to URL story, I am unable to run tests on my system. I am looking to contact team Servo for help with this.

Coder206:
@paulrouget The Servo repo is now converted from URI to URL, might be a good time to accept this PR, if it meets to all your requirements.

paulrouget:
r=me

Coder206:
r=paulrouget

PR #1093, opened 2016-06-20
Uncomment flow type annotations

tschneidereit:
The comment tokens around type annotations make the code a lot more noisy. Let's get rid of them for now - we can always use something like [recast](https://github.com/benjamn/recast) to re-comment them later.

This transformation was done using a regexp (`/ ?/\*\s*:([\w.?<>\[\], ]+)\s*\*//`), so it's bound not to have caught all instances, nor worked flawlessly. Spot-checking showed pretty good results, though, and things still compile, flow-check, and work.

I had to fix a few issues in `src/common/unknown.js` and `src/common/style.js` to make things compile again.

r? @Gozala.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1093)


Gozala:
## General remarks

1. I noticed in most cases white space is introduced after colon (for example `output: Output.Model`) my preference would be not to `output:Output.Model`, especially for return types. That being I don't care enough to block this or to even argue about it.
2. There are tons of flow comment blocks like  `/*:: ... */` where we usually:
   1. Import external type definitions
   2. Define types definitions
   3. Describe class property types
       I believe babel requires an extra [transform-class-properties](http://babeljs.io/docs/plugins/transform-class-properties/) *(could be different one)* plugin (that we don't currently use) to handle it.

  If we go with flow syntax extensions we should go all the way. Although doing rest in a followup change maybe a fine way to go about it.

tschneidereit:
Right, I didn't know how to handle the comment blocks. Thanks for pointing me to the plugin, I'll play around with that tomorrow.

As for the space after `:`, that is gone in the next PR, as the formatter I used can't be configured to keep it :) I'll re-run my changes to remove it for this PR, though.

src/common/unknown.js
```diff
@@ -10,27 +10,26 @@ import type {Action} from "./unknown";
...
     console.error(...params);
   });
+}
 
 
-export const update = /*::<model, action>*/
-  (model/*:model*/, action/*:action*/)/*:[model, Effects<action>]*/ => {
-    console.warn('Unknown action was passed & ignored: ', action, Error().stack);
-    return [model, Effects.none];
-  };
```
Gozala: Just curios what was the issue with:

```js
export const update = <model, action>
  (model:model, action:action):[model, Effects<action>] => {
    console.warn('Unknown action was passed & ignored: ', action, Error().stack);
    return [model, Effects.none];
  };
```

Asking because it supposed to and if it does not maybe a babel or our configuration issue & if former I'd like to submit a bug.

Gozala:
Alright I went over the changes (not very carefully) and they seem all right. Only reason I have not merged it yet because of the quoted commend below:

> As for the space after :, that is gone in the next PR, as the formatter I used can't be configured to keep it :) I'll re-run my changes to remove it for this PR, though.

I am not sure if you want to land this first & then remove spaces or remove spaced and then land. So I'll let you decide hen to land what ;)

r+

tschneidereit:
Ok, the comment blocks are indeed fine with the plugin pointed out in the first comment. What's still not fine is using generics on arrow functions. Consider [this testcase](http://babeljs.io/repl/#?evaluate=true&lineWrap=true&presets=es2015%2Creact%2Cstage-2&code=%0Aexport%20const%20send%20%3D%20%3CM%3E(message%3AM)%3ATask%3CNever%2C%20void%3E%20%3D%3Enew%20Task(succeed%20%3D%3E%20%7B%0A%20%20%20%20%20%20window.dispatchEvent(new%20window.CustomEvent(%22mozContentEvent%22%2C%20%7B%0A%20%20%20%20%20%20%20%20bubbles%3A%20true%2C%20cancelable%3A%20false%2C%20detail%3A%20message%0A%20%20%20%20%20%20%7D))%3B%0A%0A%20%20%20%20%20%20succeed(void(0))%3B%0A%20%20%20%20%7D)%3B): `flow` itself is perfectly happy with it, but the babel plugins don't like any part of it.

Apart from fixing the babel bug, I see two ways forward:
1. keep these as comments
2. change all the affected things from arrow functions to normal ones

Long-term I would be in favor of 2), but to get things landed here, I'll try my hand at some more-complicated regexp to do 1) as painlessly as possible.

tschneidereit:
Ah, I found @Gozala's bug for the arrow functions thing: https://phabricator.babeljs.io/T7330

tschneidereit:
@Gozala, I substantially updated the patches, and would appreciate if you could take another look. If you think this works, please just merge it. If not, I'll continue tweaking.

src/common/cursor.js
```diff
@@ -7,13 +7,13 @@
...
-/*::
+
 import type {Cursor} from "./cursor"
 export type {Cursor}
-*/
+
 
 export const cursor = /*::<from, to, in, out>*/
-  (config/*:Cursor*/)/*:(model:from, action:in) => [from, Effects<out>]*/ => {
+  (config:Cursor)/*:(model:from, action:in) => [from, Effects<out>]*/ => {
```
Gozala: Any reason to keep return type annotation as comment ? If it breaks babel or flow or something else we should report a bug.

src/common/devtools.js
```diff
@@ -148,7 +148,7 @@ const report =
 
 
 export const init =
-  ({isActive}/*:{isActive:boolean}*/)/*:[Model, Effects<Action>]*/ => {
+  ({isActive}/*:{isActive:boolean}*/):[Model, Effects<Action>] => {
```
Gozala: Argument type here also seems to be commented, likely because transformation script did not considered annotations with anonymous types.

src/common/favicon.js
```diff
@@ -28,7 +28,7 @@ const FAVICON_SIZE = 16 * window.devicePixelRatio;
  * }
  */
 export const getBestIcon =
-  (icons/*:Array<Icon>*/)/*:{ bestIcon: ?Icon, faviconURI: ?URI}*/ => {
+  (icons:Array<Icon>)/*:{ bestIcon: ?Icon, faviconURI: ?URI}*/ => {
```
Gozala: Return type is still commented here, presumably for the anonymous type used in the annotation.

src/common/history.js
```diff
@@ -6,16 +6,16 @@
...
 
-/*::
+
 import type {Integer} from "../common/prelude"
-*/
+
 
 export const readTitle = /*::<value, model:{title?:string}>*/
-  (model/*:model*/, fallback/*:value*/)/*: string | value*/ =>
+  (model:model, fallback:value)/*: string | value*/ =>
```
Gozala: return type is still in comment.

src/common/image.js
```diff
@@ -23,11 +23,11 @@ const baseStyleSheet/*:StyleSheet*/ = Style.createSheet
   );
 
 export const view =
-  (key/*:string*/, styleSheet/*:StyleSheet*/)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
-  ( model/*:Model*/
-  , address/*:Address<Action>*/
+  (key:string, styleSheet:StyleSheet)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
+  ( model:Model
+  , address:Address<Action>
   , contextStyle/*?:ContextStyle*/
```
Gozala: Type annotation here is still commented, presumably because script does not considers optional argument syntax - `contextStyle?:ContextStyle`.

src/common/image.js
```diff
@@ -23,11 +23,11 @@ const baseStyleSheet/*:StyleSheet*/ = Style.createSheet
   );
 
 export const view =
-  (key/*:string*/, styleSheet/*:StyleSheet*/)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
-  ( model/*:Model*/
-  , address/*:Address<Action>*/
+  (key:string, styleSheet:StyleSheet)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
```
Gozala: Return type annotation is still commented out (I think it may actually no longer be necessary with latest version of flow).

src/common/keyboard.js
```diff
@@ -92,7 +92,7 @@ if (!('key' in window.KeyboardEvent.prototype)) {
   ( window.KeyboardEvent.prototype
   , 'key'
   , { get: getKey
-    /*::, value: void(0)*/
+    , value: void(0)
```
Gozala: I don't seem to recall a reason properly, but I think only reason `value` was here is because flow was incorrectly expecting `value` even though getter was provided. I don't think uncommenting changes anything, but I'd rather avoid it in case there is something I'm missing to account for.

src/common/keyboard.js
```diff
@@ -158,7 +158,7 @@ const writeChord = event => {
 
 
 export const bindings = /*::<Action>*/
-  (bindingTable/*:BindingTable<Action>*/)/*:(event:KeyboardEvent) => Action | Abort*/ => {
+  (bindingTable:BindingTable<Action>)/*:(event:KeyboardEvent) => Action | Abort*/ => {
```
Gozala: Return type is still commented.

src/common/performance.js.flow
```diff
@@ -1,3 +0,0 @@
-declare export class performance {
```
Gozala: Why do we change `.js.flow` to `.js` here ? It's definitely not JS and flow given `.js.flow` a precedence over `.js` during imports.

src/common/prelude.js
```diff
@@ -112,7 +112,7 @@ const Null = () => null;
 // @FlowIssue: Frow is unable to infer
 const Void = () => void(0);
 
-export const always = /*::<a>*/(a/*:a*/)/*:(...args:Array<any>)=>a*/ => {
+export const always = /*::<a>*/(a:a)/*:(...args:Array<any>)=>a*/ => {
```
Gozala: Return type is commented out.

src/common/prelude.js
```diff
@@ -163,10 +163,10 @@ export const batch = /*:: <model, action>*/
 }
 
 export const tag = /*::<tag:string, kind>*/
-  (tag/*:tag*/)/*:(value:kind) => Tagged<tag, kind>*/ =>
+  (tag:tag)/*:(value:kind) => Tagged<tag, kind>*/ =>
```
Gozala: Return type is commented out.

src/common/request-animation-frame.js
```diff
@@ -16,7 +16,7 @@ const state =
 const frameTime = 1000 / 60
 
 export const requestAnimationFrame =
-  (request/*:(time:number) => any*/)/*:number*/ => {
+  (request/*:(time:number) => any*/):number => {
```
Gozala: request type is commented out.

src/common/style.js
```diff
@@ -61,7 +61,7 @@ export const mix = (...styles/*:Array<?Rules>*/)/*:Rules*/ => {
   }
 
   const composedStyle = id !== null ?
-    composedStyles[id/*::.toString()*/] :
+    composedStyles[id.toString()] :
```
Gozala: This was actually commented out for a reason, should have added comment explaining. `id` in this case is a `symbol` but flow does not yet support computed property access via non string `id` so this is a hack to make flow thing we're looking up via string rather than symbol.

src/common/style.js
```diff
@@ -71,7 +71,7 @@ export const mix = (...styles/*:Array<?Rules>*/)/*:Rules*/ => {
     const composedStyle = Object.assign({}, ...styles);
     // @FlowIssue: Flow does not get spread here.
     composedStyle[ID] = id;
-    composedStyles[id/*::.toString()*/] = composedStyle;
+    composedStyles[id.toString()] = composedStyle;
```
Gozala: Same deal as in https://github.com/browserhtml/browserhtml/pull/1093/files#r68305288

src/common/toggle.js
```diff
@@ -102,11 +102,11 @@ export const update =
 
 
 export const view =
-  (key/*:string*/, styleSheet/*:StyleSheet*/)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
-  ( model/*:Model*/
-  , address/*:Address<Action>*/
+  (key:string, styleSheet:StyleSheet)/*:(model:Model, address:Address<Action>, contextStyle?:ContextStyle) => DOM*/ =>
```
Gozala: Return type is still commented.

src/common/toggle.js
```diff
@@ -102,11 +102,11 @@ export const update =
...
+  ( model:Model
+  , address:Address<Action>
   , contextStyle/*?:ContextStyle*/
```
Gozala: Optional argument is still commented.

Gozala:
> Ah, I found @Gozala's bug for the arrow functions thing: https://phabricator.babeljs.io/T7330

Oh right I forgot about it. I talked to flow core team today (they are also contributing to babel it seems) and they are looking into this, so hopefully there's going to be fix sometime soon.

Gozala:
> @Gozala, I substantially updated the patches, and would appreciate if you could take another look. If you think this works, please just merge it. If not, I'll continue tweaking.

Alright I did another pass over it & provided some feedback. I am afraid there are some issues that needs to be addressed before we can land, like some flow comment blocks supposed to trick flow and not actually run or will alter behavior & may introduce bugs). Which those and dep version comment addressed r+.

I also pointed out some commented annotations, but I'm fine with leaving them for now if it's tricky to uncomment those, it's small enough number of them that we could manually uncomment them as a followup. Or if you want to deal with them before landing that's fine by me as well.

Gozala:
@tschneidereit There are still two pending remarks from me:

[We trick flow & should no uncomment unless there's good reason](https://github.com/browserhtml/browserhtml/pull/1093#discussion_r68295293)
[I'm not sure why file was renamed](https://github.com/browserhtml/browserhtml/pull/1093#discussion_r68295645)

Gozala:
@tschneidereit I have bugged right ppl and latest version of babel has a fixe for polymorphic arrow functions. I'll submit a pull request to update our babel version to get that resolved. And with that we should be able to uncomment only remaining annotations & avoid rewriting arrow functions as an old school ones.

src/common/keyboard.js
tschneidereit: This is just silly: flow wants the object literal to have a `value` field for no good reason whatsoever. I re-added the comment, with an explanation.

src/common/performance.js.flow
tschneidereit: Because otherwise this happens during bundling:
`Failed to browserify: 'main' Cannot find module './performance' from '/Users/till/mozilla/servo/browserhtml/src/common'`

With the comments around flow types removed everywhere, this file is as much JS as everything else, so I don't see a reason not to do the renaming. In fact, I think we should just fold the contents of the `.flow` files into the respective `.js` files, but that's something for another PR.

src/common/keyboard.js
Gozala: @tschneidereit I think problem is actually in `Object.defineProperty` API being non typing friendly. It means that descriptor type would have to be:

```js
type PropertyDescriptor <x> = 
  | { value: x }
  | { get: () => x }
  | { set: (x) => any }
  | { get: () => x, set: () => x }
```

And such union types are difficult to support. At least flow does not support them. Alternative would be to define descriptor type as:

```js
type PropertyDescriptor <x> = {
  value?: x,
  get?: () => x,
  set?: (x) => any
}
```

In which case flow won't really be able to tell you if you're making mistake and provide no properties at all or maybe `value` and `set`. 

My guess is they resorted to just making all of the properties mandatory.

tschneidereit: @Gozala, well but this solution is definitely the worst of all worlds: `Object.defineProperty` throws if you specify both `value` and either `get` or `set` (or both). While the enum/variant type would be best, the version with all values optional would be better than what they have now.

PR #1094, opened 2016-06-20
Mass-reformat all JS files - WIP

tschneidereit:
This was done using WebStorm, with the formatting settings slightly tweaked from the defaults. I'm not 100% happy with the results, but what's good about this formatter is that, in difference to the others I tried, it normalizes line breaks and blank lines. Some quick testing shows that other formatters work quite a bit better after this one, so perhaps it's a good intermediate step.

Note that I don't think we should land this as-is - I'll experiment more with applying further reformattings that we should then be able to run automatically and that our integration tests could verify.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1094)

### DIFF
--- a/src/about/newtab/newtab/tile.js
+++ b/src/about/newtab/newtab/tile.js
@@ -4,97 +4,80 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import {html, Effects} from "reflex";
-import * as Style from "../../../common/style";
-import * as Unknown from "../../../common/unknown";
+import { html, Effects } from 'reflex'
+import * as Style from '../../../common/style'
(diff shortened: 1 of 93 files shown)


src/about/newtab/newtab/tile.js
```diff
@@ -4,97 +4,80 @@
...
-            : styleSheet.titleLight
-            )
-          )
-        }
-      , [ model.title
-        ]
-      )
-    ]
-  );
+export const view = (model:Model, address:Address<Action>, isDark:boolean):DOM =>html.a({
```
Gozala: Please let's not do this :)

tschneidereit: Yeah, I fully agree. To be able to work with automatic formatters we'll have to change code like this, though: without doing something like assigning intermediary results or, in this case, extracting type definitions, a formatter will always be too dumb to properly format this.

PR #1097, opened 2016-06-23
Optimize imports

tschneidereit:
WebStorm has this nice feature for optimizing all imports, and it happens to work quite splendidly on our codebase. Substantially cuts down on LOC, and at least for me has a more visually pleasing result.

Applies on top of #1093, and is quite straight-forward.

r? @Gozala

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1097)

### DIFF
--- a/src/common/cursor.js
+++ b/src/common/cursor.js
@@ -7,10 +7,10 @@
 
 import {Effects} from "reflex";
 
-/*::
+
 import type {Cursor} from "./cursor"
 export type {Cursor}
-*/
+
 
 export const cursor = /*::<from, to, in, out>*/
   (config/*:Cursor*/)/*:(model:from, action:in) => [from, Effects<out>]*/ => {
--- a/src/common/cursor.js
+++ b/src/common/cursor.js
@@ -8,28 +8,27 @@
 import {Effects} from "reflex";
 
 
-import type {Cursor} from "./cursor"
+import type {Cursor} from './cursor'
 export type {Cursor}
 
 
-export const cursor = /*::<from, to, in, out>*/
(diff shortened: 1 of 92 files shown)


Gozala:
@tschneidereit last change causes type checker to fail, looking at the log it seems that `Effects` is no longer imported in `cursor.js` presumably because WebStorm ignores flow type annotations and there for strips imports that are only used for annotations.

src/common/cursor.js
```diff
@@ -5,15 +5,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 
-import {Effects} from "reflex";
```
Gozala: Removing this import causes flow to fail, as type annotation below (on like 13) refers to `Effects` that is no longer in scope.

I imagine two reasons why web storm is deciding to remove this import:

1. Because it does not actually understands type annotations so assumes that imports are redundant.
2. Because `Effects` is referred from comment web storm ignores it and assumes import is redundant.

I think we could actually use `import type {Effects} from "reflex"` as we only refer to `Effects` from the type position. Which I presume web storm won't remove and ignore. If on the other hand web storm is actually smart enough and No 1 isn't the issue here, then uncommenting return type annotation on line 13 is likely going to be solution for web storm to see that `Effects` is used.

P.S. Fixing this may uncover more type errors due to flow being able to analyze more if it does not fail.

tschneidereit: Bah, I did multiple runs of this with slightly different settings, and
seemingly forgot to test after the last one. Will push an update.

Gozala:
@tschneidereit there are [two pending remarks](https://github.com/browserhtml/browserhtml/pull/1093#issuecomment-228429736) in #1093

r=me for last commit, import optimizations.

PR #1102, opened 2016-06-27
Fix browser.html when using native titlebar.

Gozala:
Turns out VirtualDOM `thunk` does not like resolution to `””` and always expects node.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1102)


tschneidereit:
r=me

PR #1104, opened 2016-06-28
Add cmd-q shortcut to quit

ConnorGBrewster:
Not 100% sure of these changes, but it works :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1104)

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -23,6 +23,7 @@ import * as Easing from "eased";
 import {always, batch, tag, tagged} from "./common/prelude";
 import {cursor} from "./common/cursor";
 import {Style, StyleSheet} from './common/style';
+import type {Result} from "./common/result"
 
 import {identity, compose} from "./lang/functional";
 
@@ -77,6 +78,8 @@ export type Action =
   | { type: "Reloaded" }
   | { type: "OpenURL", uri: URI }
   | { type: "Close" }
+  | { type: "CloseRuntime" }
+  | { type: "Closed", result: Result<Error, void> }
   // @TODO: Do not use any here.
   | { type: "Modify", modify: ID, action: any }
   | { type: "Open" }
@@ -256,6 +259,12 @@ const updateSidebar = cursor({
   update: Sidebar.update
 });
 
+const closed = (model, result) =>
+  ( result.isOk
+   ? [ model, Effects.none ]
+   : [ model, Effects.perform(Unknown.error(result.error)) ]
+  );
+
 const Reloaded:Action =
   { type: "Reloaded"
   };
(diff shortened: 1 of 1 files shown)


src/browser.js
```diff
@@ -261,6 +261,12 @@ const updateSidebar = cursor({
   update: Sidebar.update
 });
 
+const closed = (model, result) =>
+  ( result.isOk
+  ? [ model, Effects.none ]
+  : [ model, Effects.perform(Unknown.error(result.error)) ]
```
Gozala: Flow will infer return type of the `closed` function to be `[Model, Tagged<"Errored", void>]` where second item in the tuple will end up being send into `update`. But then update does not actually supports `Action` of that type, which is why type checker fails. On possible fix is to do `.map(NoOp)` as seen on line 692. But honestly our `Unknown.error` function should be updated to make this unnecessary.

Gozala:
Thanks @ConnorGBrewster for contributing! I am afraid your changes don't type check. After looking into  errors it seems that problem is with [returned type](https://github.com/browserhtml/browserhtml/pull/1104/files#r68850748). That being said the fact that requires extra hoops is just odd and stupid. There for I wrote a patch to fix this (see #1109). Once that lands, if you just rebase / merge master type checker will be happy & we'll be happy to merge this inn.

Gozala:
@ConnorGBrewster BTW I have being writing docs about architecture used in this project https://gozala.gitbooks.io/reflex/content/ that may be helpful in understanding code. I would also very much appreciate feedback if you'll end up reading it.

ConnorGBrewster:
@Gozala alright I rebased against those changes. I will definitely read though your docs, I have been curious about how bhtml is structured.

ConnorGBrewster:
Hmm still fails, I will look into this more in a bit.

Gozala:
@ConnorGBrewster ooh there is one more thing that I missed, sorry for that. This time type checker fails because you have introduced new action types, that have not being added to the `Action` type definition here: https://github.com/ConnorGBrewster/browserhtml/blob/97f455c4cee5e02dd7aa8894c46008d47b9dedcd/src/browser.js#L36-L83

BTW you can run type checker locally  by running `npm test`.

From what I can tell there are two types are missing in the `Action` union:

```js
| { type: "CloseRuntime" }
| { type: "Closed", result: Result<Error, void> }
```

And you'd need to also `import type {Result} from "./common/result"`.

ConnorGBrewster:
Alright tests passed locally! should be good to go now

Gozala:
@ConnorGBrewster awesome! Thanks for the patch & patience of getting through all the hoops!

PR #1109, opened 2016-06-28
Make task returned by Unknown.log etc. polymorphic

Gozala:
Tasks used to have static success type that made no sense:

1. Because those task never actually succeed or fail.
2. Because it required users to either support success type as action or map it to some action.

With this change success type is type parameter, which means that it is only going to be restricted by further inference.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1109)


tschneidereit:
I like how this is mostly code removal.

PR #1110, opened 2016-06-29
Skip New Tab page when switching tabs with ctrl+tab

tschneidereit:
r? @gozala

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1110)


Gozala:
Current behavior is actually intentional UX choice. Maybe let's discuss this afterwards ?

PR #1111, opened 2016-06-29
Remove redundant .js.flow files

Gozala:
Basically type annotations moved inside the implementations to reduce surface of errors where things can get out of sync.

Note: Some renaming became necessary due to following issues facebook/flow#1890

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1111)


tschneidereit:
I skimmed this, only looking at the renaming parts a bit closer. Looks good to me.

PR #1113, opened 2016-06-29
Switch to HTML version of DDG.

### LINKED ISSUE #1103: Use https://duckduckgo.com/html/?q=foobar for the search page
We have issues with duckduckgo, but the non-js version of ddg appears to work.

https://duckduckgo.com/html/?q=foobar

Gozala:
Fixes #1103

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1113)


tschneidereit:
r=me

PR #1125, opened 2016-06-30
Increase default vertical size to 740

tschneidereit:
Makes github work by changing a media query's result.

r? @paulrouget

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1125)


paulrouget:
r=me

PR #1132, opened 2016-06-30
Bundle servo demos.

### LINKED ISSUE #1131: Bundle New Tab page demos
The four links at the bottom currently target servo.org. We should instead bundle them to avoid network issues from making the experience worse than it should be.

Gozala:
Fixes #1131

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1132)


tschneidereit:
r=me assuming @metajack's Windows path length issues aren't a blocker for this.

metajack:
@tschneidereit those are just warnings, so not a blocker at all.

PR #1135, opened 2016-06-30
Swap menu demo for spinning demo

ConnorGBrewster:
As per servo/servo#11941 we are switching to the spinning demo instead of the menu demo.

Note the spinning demo still has some issues, but hopefully are easier to fix than the issues with the menu demo. servo/servo#11759

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1135)


asajeffrey:
Looks good to me. @tschneidereit? @paulrouget?

Gozala:
r=me

PR #1139, opened 2016-06-30
Switch spinning demo to evil ad demo

ConnorGBrewster:
Switching because spinning has issues (servo/servo#11970) and evil ad is much cooler.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1139)


src/about/newtab/tiles.json
```diff
@@ -70,9 +70,9 @@
     , "src": "./tiles/demo-transparent_rects.png"
     }
   , "11":
-    { "title": "Demo: Spinning"
-    , "uri": "./demos/spinning.html"
-    , "src": "./tiles/demo-spinning.png"
+    { "title": "Demo: Evil Ad"
+    , "uri": "./demos/evil_ad/physics/with-ad.html"
```
Gozala: This points to a file that does not exists. Can you please fix and resubmit ?
I had to revert this change.

ConnorGBrewster: that odd, somehow this was working on my local machine 😕

PR #1141, opened 2016-06-30
Switch spinning demo to evil ad demo

ConnorGBrewster:
@Gozala sorry about the previous PR, this should have the evil ad files structured properly. (The uri was fine, files were in the wrong spot).

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1141)


tschneidereit:
@ConnorGBrewster the tile image is quite large, is that intentional? I'd much prefer an image that exactly fits the dimensions of the tile. Otherwise we waste file size, plus an offline resizing should yield better visual results.

ConnorGBrewster:
@tschneidereit it is roughly the same size as the others. Its dimensions are `794x480` and most the others are around `800x500`

tschneidereit:
@ConnorGBrewster oh, I didn't realize that the other tiles are that large, too. Never mind, then: dpi really doesn't matter in this context, having it match the other tiles' pixel dimensions makes the most sense.

I'd r+, except I think we should also remove the other demo's files: no need to package them up if nobody'll ever see them.

Gozala:
> @ConnorGBrewster oh, I didn't realize that the other tiles are that large, too. Never mind, then: dpi really doesn't matter in this context, having it match the other tiles' pixel dimensions makes the most sense.

I believe other tiles are `800 x 500` except demo ones, we should probably cut them as such as well.

@ConnorGBrewster @metajack also asked to place "evil ad" demo to the left bottom corner. Could you please do that as well so we can land it in one merge ?

ConnorGBrewster:
@Gozala moved the demo to bottom left corner and trimmed all the demo tiles to 800x500 @ 72ppi

Gozala:
Awesome thanks

PR #1142, opened 2016-06-30
Copy evil ad js to dist

ConnorGBrewster:
The js files with the evil ad demo were not being copied by gulp.

Hopefully this will be the last fixup here :wink:

Note: copying all `*.js` broke the newtab page, so I just have this copying the js for the evil ad demo only.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1142)


tschneidereit:
r=me

PR #1143, opened 2016-06-30
Set rect count in Transparent Rects demo to a saner value

tschneidereit:
Who'd ever need more than 500 rects?

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1143)


tschneidereit:
r=me

PR #1166, opened 2016-07-05
Fix input updates so that content does net get's messed up when typing quickly 

### LINKED ISSUE #1157: Suggestions sometimes blow away what you typed
@fitzgen keeps running into issue where what he types gets blown away by the suggestions. Typically typing `git` seems to reproduce it every now and then.

Gozala:
There were set of issues related to the input field:

- [x] Sometimes response from search service would arrive when input already had changed. Updates to suggestions or input would blow away data entered in the meantime.
- [x] It was possible that input would update while message from `Search` service delivered to an input, since messages are delivered async. There for it was possible that suggestion would become obsolete.
- [x] Selection changes were applied async. Intent was to apply them on next tick, while on servo it would happen on `setTimeout(cb, 0)` due to no native `Promise` implementation. This caused cursor position to get out of sync when typing quickly.

Fixes #1157

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1166)

### DIFF
--- a/src/browser/Navigators/Navigator.js
+++ b/src/browser/Navigators/Navigator.js
@@ -712,12 +712,7 @@ const suggest =
   (model, suggestion) =>
   updateInput
   ( model
-  , Input.Suggest
-    ( { query: model.assistant.query
-      , match: suggestion.match
-      , hint: suggestion.hint
-      }
-    )
+  , Input.Suggest(suggestion)
   )
 
 const activateAssistant =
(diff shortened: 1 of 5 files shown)


Gozala:
As all of my team is on vacation, I'll be asking @fitzgen (who reported the issue) to do the review, so it can ship with nightlies soon.

fitzgen:
As an outsider to this code base, this looks good to me and the approach seems sound.

Will try and manually verify whether I can still reproduce and report back.

fitzgen:
Could no longer reproduce the suggestions getting blown away.

PR #1170, opened 2016-07-06
Stop querying suggestions on delete.

### LINKED ISSUE #1167: Auto complete should not provide selected completions on deletes
Currently if you type something, very likely you'll get an autocomplete with completion being selected. Idea is that you can continue typing, if competition isn't what you care about, or hit enter to navigate to suggestion.

Unfortunately if you had a typo and need to delete several characters, you'll end up needing to double backspace per character.

1. To delete selection.
2. To delete actual character.

This is not ideal. We could probably do what Safari does here. Which is do not provide selected completion if backspace was typed, only provide it if additional characters were typed.

Gozala:
Fixes #1167

With this change input changes do not query suggestion service if new value is included by a previous one, as we do not want to provide completions while user is deleting text. Once new character is typed new suggestions / completions will provided.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1170)

### DIFF
--- a/src/browser/Navigators/Navigator/Input.js
+++ b/src/browser/Navigators/Navigator/Input.js
@@ -179,7 +179,7 @@ export const update =
       case 'Editable':
         return mapFX(EditableAction, Editable.update(model, action.editable));
       case 'Change':
-        return mapFX(EditableAction, Editable.update(model, Editable.Change(action.value, action.selection)));
+        return change(model, action.value, action.selection);
       case 'Show':
         return [merge(model, {isVisible: true}), Effects.none];
       case 'Hide':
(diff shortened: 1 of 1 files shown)


src/browser/Navigators/Navigator/Input.js
```diff
@@ -197,6 +197,26 @@ export const update =
     }
   };
 
+const change =
+  (model, value, selection) => {
+    const [edit, editFX] =
+      Editable.update(model, Editable.Change(value, selection));
+
+    const fx =
+      ( model.value.includes(edit.value)
```
jsantell: Wouldn't this be `edit.value.includes(model.value)`? If `edit.value` is currently "hell" and the model was previously "hello", etc., unless I got these values backwards

jsantell:
One comment, r+ if that looks good

src/browser/Navigators/Navigator/Input.js
Gozala: > Wouldn't this be edit.value.includes(model.value)? If edit.value is currently "hell" and the model was previously "hello", etc., unless I got these values backwards

So if `edit.value === "hell"` and `model.value === "hello"` this means that `o` was deleted. Theres for `model.value.includes(edit.value) === true` since `"hello".includes("hell")`.

Does that makes sense ?

jsantell: Ah yes, got it backwards in my head, looks good!

Issue #930, opened 2016-03-15
Need to type backspace twice to delete a character in the url bar


laurentj:
To delete one character in the url bar, with the backspace key, I need to type two times on this key. 

Servo build: commit 075ce98 , release build, on linux Kubuntu 15.04 64bits, azerty keyboard.

paulrouget:
Can reproduce on Mac, Qwerty. It appears to be browserhtml specific.

paulrouget:
Might be related to the recent implementation of `setSelectionRange`.

paulrouget:
Can someone investigate and see if it's a Servo bug or a bhtml bug?

laurentj:
Well the bug appears only in the url bar of bhtml. Not in any other input field of web pages.

paulrouget:
But we don't have the issue with Gecko. Might be an issue with key events + `setSelectionRange`.

mbrubeck:
This is a Servo bug: servo/servo#10182.

paulrouget:
Works now.

PR #1174, opened 2016-07-07
Fix regression that prevents suggestion to update autocomplete.

### LINKED ISSUE #1173: Selecting next / previous suggestions does not update autocomplete
This regression was intruded by #1157 

`control p / control n` or hovering suggestions in drop down should update autocomplete selection in the input field.

Gozala:
Fixes #1173

Change 8acb7fd started asserting search query to input value to only update input if query matched the value. This did not take into the account that suggested autocomplete is also part of value and there for changing suggestions would no longer update autocomplete.

This change amends original assertion and compares query only to unselected fragment of value so that completion selections do not affect the assertion.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1174)

### DIFF
--- a/src/browser/Navigators/Navigator/Input.js
+++ b/src/browser/Navigators/Navigator/Input.js
@@ -140,7 +140,7 @@ export const init =
   ];
 
 const suggest = (model, {query, match, hint}) =>
-  ( model.value !== query
+  ( getUnselectedValue(model) !== query
   ? [model, Effects.none]
   : enterSelectionRange
     ( model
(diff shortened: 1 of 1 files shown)


Gozala:
@fitzgen can you please review this change as you reviewed the code that introduced regression ? Also could you please try reproducing the original input blow out issues with this change, just to make sure it's not going to bring it back.

jsantell:
lgtm

Gozala:
I just found one edge case that, I would like to also include with this fix. Related to a #930 

If you you say type **am** you'll get following in the input field (non bold part is completion selection): **am**azon. Then if you type `a` again you'll wind up with: **ama**.

Selection disappears and no suggestion will be updated. That is because we check if the new input is part of the old input and if so, we do not query search engine. I think most expected behavior would be to wind up with **ama**zon instead. Not sure about querying the search engine. On one side it still makes sense to do it as other suggestions may no longer be appropriate. On the other hand there is the risk that top suggestion could end up different.

Gozala:
Actually ignore my last comment as I have create separate issues #1175 to deal with this. It is not really related to this other than being issue with a same component.

PR #1178, opened 2016-07-07
Fix bug in selection update code.

### LINKED ISSUE #1176: Selection changes don't always reflect properly
When you type **ama** in input field you'll get bunch of suggestions when selecting through them, sometimes selection goes away instead of updating properly. See attached gif for better description

selection-updates

As you can notice once "amazon books" is selected selection disappears.

Gozala:
Fixes #1176

Previous code was trying to avoid infinite selection update loop by comparing `previous` and `next` selection ranges and dropping updates if they were equal. This did not properly handled a case where input was updated with different value `value` but equal `selection` as selection update code would run before `select` event would occur. With this change selection change now is determined by comparing `next` selection range to an actual selection on the DOM element, avoiding both infinite loops and above mentioned issue.

This change also makes `selection` field mandatory on models as `null` selection makes things less deterministic and opens possibility for race conditions.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1178)

### DIFF
--- a/src/browser/Navigators/Navigator/Input.js
+++ b/src/browser/Navigators/Navigator/Input.js
@@ -28,7 +28,7 @@ export type Model =
   { isVisible: boolean
   , isFocused: boolean
   , value: string
-  , selection: ?Editable.Selection
+  , selection: Editable.Selection
   }
 
 export type Suggestion =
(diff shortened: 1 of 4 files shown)


Gozala:
@jsantell for the reference, `selection` meta property is a hook that is triggered on [componentDidMount](https://facebook.github.io/react/docs/component-specs.html#mounting-componentdidmount) and [componentDidUpdate](https://facebook.github.io/react/docs/component-specs.html#updating-componentdidupdate) methods and are used to update DOM node's things that aren't handled by Virtual DOM abstraction.

jsantell:
r+

PR #1180, opened 2016-07-08
Improve type annotation of merge to catch more type mismatches.

Gozala:
With this changes to `merge` flow ensures that all fields of `patch` are type compatible with corresponding fields of the `model` being patched.

Second commit fixes all the mismatches made apparent by this change.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1180)


jsantell:
r+

PR #1181, opened 2016-07-08
Fix type errors introduced by #1180 over #1178

Gozala:
In #1180 type annotations of `merge` got improved to handle type mismatches. But #1178 introduced type mismatch that flow was unable to catch until #1180 fixed that. There for landing #1180 start reporting type mismatches.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1181)


jsantell:
r+

PR #1183, opened 2016-07-12
Update to flow@0.28

Gozala:
- Update reflex-virtual-driver that failed on flow@0.28
- Use `toString()` in template string variables with non strings as required in flow@0.28.
- Add missing type annotations.
- Get rid off @FlowIssue suppression comments where possible.

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1183)


jsantell:
LGTM

Issue #1179, opened 2016-07-07
Electron app doesn't work without Node integration


freebroccolo:
I am trying to load the electron app with electron `v1.2.6`. The window shows up with the address bar and you can type stuff in but it won't actually load any URLs or anything unless `nodeIntegration` is re-enabled [here](https://github.com/browserhtml/browserhtml/blob/master/main.js#L56).

Is this expected? Is there some other way to start the app with electron that doesn't have this issue?

Gozala:
Hi @freebroccolo thanks for letting us know. I have being testing it with electron v1.0.1 without `nodeIntegartion` & it works fine.

I guess something has changed in newer versions of Electron. I'll try to investigate into this, when I get a chance. That being said, Electron support isn't very high priority so I may not get to this soon.

If you'd be willing to look into what's breaking I'd be happy to help figuring / fixing things up to make so it works with latest version without needing to enable `nodeIntegration`.

Gozala:
It seems that in 1.11.1 `webview` tag is disabled when `nodeIntegration` is off in host window, which would explain this issue: https://github.com/electron/electron/releases/tag/v1.1.1

freebroccolo:
Ah, thanks for looking into this. I suspected it was something like that. Do you have a workaround in mind? What about leaving `nodeIntegration` enabled but running a preload script to remove unsafe modules?

Gozala:
> Ah, thanks for looking into this. I suspected it was something like that. Do you have a workaround in mind? What about leaving nodeIntegration enabled but running a preload script to remove unsafe modules?

That could be an intermediate solution (I'd be happy to take a pull request for this 😉).

In longer term I would like to find out why Electron disabled `webview` tag (my guess after skimming through the linked issues in release notes, is in order to re-enable sandboxing) or what is expected usage. For example how does brave browser go about it ?

freebroccolo:
> In longer term I would like to find out why Electron disabled webview tag (my guess after skimming through the linked issues in release notes, is in order to re-enable sandboxing) or what is expected usage. For example how does brave browser go about it ?

@Gozala I'm not sure how they do it in the brave browser but from [this issue](https://github.com/electron/electron/issues/6157) it sounds like preload may be the expected approach.

poiru:
[Issue #5173](https://github.com/electron/electron/issues/5173) has some context. Enabling `nodeIntegration` is not safe because of that bypass.

paulrouget:
[...]

PR #1184, opened 2016-07-12
Re-enable nodeIntegration; use electron preload

freebroccolo:
Per the discussion in #1179, this re-enables `nodeIntegration` and runs a `preload` script to delete the node modules (roughly following what happens [here](https://github.com/electron/electron/blob/9f0fc96025fec8948eeb8ca45184891c3d8e485d/lib/renderer/init.js#L85)).

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1184)

### DIFF
--- a/electron-preload.js
+++ b/electron-preload.js
@@ -1,0 +1,9 @@
+window.onerror = null;
+delete global.__dirname;
+delete global.__filename;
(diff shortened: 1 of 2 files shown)


Gozala:
Thanks!

PR #1185, opened 2016-07-12
Navigator input rewrite

Gozala:
This is a huge patch that I'll be refactoring into smaller ones and submitting them separately:

- [x] Factor out nofx #1193
- [x] Implement `appendFX(fx2, [model, fx])` helper function #1196
  - [x] Implement `port(decodeEventAsAction)` helper function #1197
    - [x] Implement `anotate(port, tag)` helper function #1198 
- [x] File renaming #1195
  - [x] Factor out `Search` and `History` IO stuff #1194
    - [x] Simplify `Suggestion` view code to avoid passing extra context info #1199
- [x] Make `Editable` and `Focusable` embeddable #1200
  - [x] Refactor `TextInput` to embed `Focusable`, `Editable`, `Control` instead of mixing them in. #1201
    - [x] Refactor `Input` to embed `Focusable`, `Editable` instead of mixing them in. #1202
      - [x] Refactor `Toggle` to embed `Button` and avoid all mixins. #1203
        - [x] Refactor `Button` to embed `Focusable` & `Control`. #1204
          - [x] Make `Focusable` monomorphic #1205 (includes #1198)
            - [x] Make `Control` monomorphic #1206
              - [x] Make `Target` monomorphic #1207
                - [x] Make `Editable` monomorphic #1208
- [x] Rename `Press` to `Click` #1210
- [x] Create generic `Selector` component decoupled from options #1211

After all above changes have landed this pull will be rebased / reset to just include changes that alter behavior of the input.

P.S.: Changes are organized in a tree structure to better visualize review queue.


Gozala:
Ok after ton of rebasing, merging and landing pulls, I merged master into this & presumably this pull should only contain relative changes now.

Gozala:
Alright after last two open pull requests will land this pull will only contain relevant changes & it would be difficult to further split it apart. In order to make review easier I'll provide overview of the changes:

- `History` and `Search` modules acted as selections themselves and `Assistant` would joggle between them, but it became difficult to manage such setup as more constrained araised.  With this change `History.js` and `Search.js` now just describe corresponding entry in the suggestions.
- `Assistant` now maintains single list of suggestions, where each suggestion either contains  `History` or `Search` entry & generic `Selector` is used to maintain selection state and deal with `view` logic.
- `Suggestion` used to sort of provide interface for different suggestion types to implement. It's new behavior is totally unrelated now. What it essentially does is provide an abstraction between different types of suggestions in this case `History` and `Search`. So it basically boxes messages and models and just proxies to the corresponding `view`, `update`, etc. functions.
- The meat of changes are in `Input` and `Assistant` modules. By which I mean actual changes in the behavior of the UI.
  - `Input` used to guess what part of it's value was typed query and what part was suggestion. Those guesses were not correct in some edge cases so instead input just keeps track of typed query in a separate `query` field now instead.
 - Again `Input` was trying to make guesses if you were deleting character, so it would not get into your way via autocomplete. As above those guesses were wrong in some edge cases so it made more sense to keep track of `deleting` state in a separate field.
 - `CancelSuggestion` message was added to `Input` so it can tell parent remove selection from suggestions.
 - Now that `Input` knows whether user is deleting or typing query it can spawn queries in more appropriate manner. 
 - `Assistant` now queries `Search` and `History` services and updates `suggestions` when results are received.
 - Update in suggestions now retains selected `suggestion` if there is one so user does no loose it due to delayed response from server.
- If new `query` matches currently selected suggestion selection is retained, otherwise it's dropped.
- Now `Query`  message of `Assistant` has additional `suggest:boolean` field that allows querying it silently or not. So when user is deleting silent `Query`s are submitted and there for assistant updates suggestions but does not provide autocomplete. 

Now in terms of behavior this is what we I was going for:

1. When you type into input suggestions are provided in the selector in form of dropdown.
2. Top suggestion is suggested as an autocomplete (selected text that follows typed query).
3. If you just type autocomplete and suggestions stay out of you way, due to just providing selections that are replaced by characters you type.
4. If selected suggestion matches new query (typed character you typed is a first selected character) then selection range just update to include new query and suggestions in the drop down update as well.
5. If selected suggestion does not match new query (typed character is not the first selected character) then selected text is replaced by new character and suggestions are updated to match new query. Once new results come back Top suggestion is provided via autocomplete.
6. There is no longer a case when typed query can be lost due to switching between suggestions, because Top suggestion is guaranteed to either contain query or otherwise typed query is provided as top suggestion as is.
7. If user deletes provided autocomplete, autocomplete (selected text after typed query) is just removed & selected suggestion gets deselected.
8. If user continues to text from input, autocomplete is not provided although suggestions in drop down are updated with that query.
9. If user starts typing again things reset to normal and suggestions work as described.
10. If all characters are deleted then suggestions go away.
11. Selection direction of auto-complete is backwards as it provides more natural expansion / subtraction of the auto-complete.
12. If something got selected while new query results were awaited, selected item is retained and moved to the top.

paulrouget:
r=me

PR #1187, opened 2016-07-16
Support loading an initial URL

### LINKED ISSUE #1089: Support loading an initial URL
This would allow us to support passing an initial URL like ./`mach run -r -b http://mozilla.org`, which I always forget doesn't work.

linclark:
Fixes #1089 

## Testing instructions

Start servo with a url query parameter appended to the url, for example:  `../mach run -r -- -b -w --resolution=1024x740 --pref dom.mozbrowser.enabled --pref dom.forcetouch.enabled --pref shell.builtin-key-shortcuts.enabled=false http://localhost:6060?url=http://example.com
`

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1187)

### DIFF
--- a/src/browser/Navigators.js
+++ b/src/browser/Navigators.js
@@ -13,6 +13,8 @@ import * as Overlay from "./Navigators/Overlay"
 import * as Navigator from "./Navigators/Navigator"
 import * as URI from "../common/url-helper";
 import * as Tabs from "./Sidebar/Tabs";
+import * as Runtime from '../common/runtime';
+
 
 
 import type {Address, DOM} from "reflex"
@@ -95,7 +97,6 @@ const Card =
   , deselect: Navigator.deselect
   }
 
-
 const tagDeck =
   (action:Deck.Action<Navigator.Action, Navigator.Flags>):Action => {
     switch (action.type) {
@@ -177,7 +178,26 @@ export const init =
       ( Card
       , deck
       , flags
-      )
+      );
+    const [deck3, $deck3] =
+      ( Runtime.env.url
+      ? Deck.open
+        ( Card
+        , deck2
+        , { output:
+            { uri: Runtime.env.url
+            , disposition: 'default'
+            , features: ''
+            , ref: null
+            , guestInstanceId: null
+            }
+          , overlay: true
+          , isPinned: true
+          , isInputEmbedded: true
+          }
+        )
+      : [deck2, $deck2]
+      );
 
     const display =
       ( shrink
@@ -188,10 +208,10 @@ export const init =
       )
 
     const [animation, $animation] = Animation.init(display);
-    const model = new Model(zoom, shrink, deck2, animation);
+    const model = new Model(zoom, shrink, deck3, animation);
     const fx = Effects.batch
       ( [ $deck.map(tagDeck)
-        , $deck2.map(tagDeck)
+        , $deck3.map(tagDeck)
         , $animation.map(tagAnimation)
         ]
       )
--- a/src/browser/Navigators.js
+++ b/src/browser/Navigators.js
@@ -185,15 +185,13 @@ export const init =
         ( Card
         , deck2
         , { output:
-            { uri: Runtime.env.url
+            { uri: URI.read(String(Runtime.env.url))
             , disposition: 'default'
+            , name: ''
             , features: ''
             , ref: null
             , guestInstanceId: null
             }
-          , overlay: true
-          , isPinned: true
-          , isInputEmbedded: true
           }
         )
       : [deck2, $deck2]


Gozala:
@linclark I am afraid your changes are failing type checker, which is why travis build has failed. I know flow error messages can be quite cryptic and it takes some time to get used to. Good news is flow team is trying to improve them ;)

I'll comment inline and quote errors from flow to help make sense out of those errors.

src/browser/Navigators.js
```diff
@@ -177,7 +178,26 @@ export const init =
...
+        , { output:
+            { uri: Runtime.env.url
+            , disposition: 'default'
+            , features: ''
+            , ref: null
+            , guestInstanceId: null
+            }
+          , overlay: true
+          , isPinned: true
+          , isInputEmbedded: true
```
Gozala: Last three options `overlay`, `isPinned` and `isInputEmbedded` they have special setting for `about:newtab` tab to make it unclosable & have address bar permanently expanded. For regular tabs those actually would be `false`-es. That being said those field are optional so I suggest to just omit those field in this case and correct defaults will be used.

src/browser/Navigators.js
```diff
@@ -177,7 +178,26 @@ export const init =
...
+      ( Runtime.env.url
+      ? Deck.open
+        ( Card
+        , deck2
+        , { output:
+            { uri: Runtime.env.url
+            , disposition: 'default'
+            , features: ''
+            , ref: null
+            , guestInstanceId: null
```
Gozala: `Navigators` is probably most complex  component as it delegates to polymorphic (a.k.a generic `Deck` component) and it may not be very clear what `Deck.open` should be passed. I'll try to explain:

So `Deck.open` has following type signature:
```js
type open = <action, model, flags>
  (api:Card<action, model, flags>, model:Model<model>, flags:flags):
  [Model<model>, Effects<Action<action, flags>>] 
```

Since `api` argument in this case is `Card` we can derive what the types of `action`, `model` and `flags` are from there. In this case [it's a `Navigator`](https://github.com/linclark/browserhtml/blob/a62789ecb28a310e51d447828bf22e1ff92d0fc1/src/browser/Navigators.js#L92-L98) and from it's [`init` function](https://github.com/linclark/browserhtml/blob/a62789ecb28a310e51d447828bf22e1ff92d0fc1/src/browser/Navigators/Navigator.js#L358) we can derive that [`Flags` type](https://github.com/linclark/browserhtml/blob/a62789ecb28a310e51d447828bf22e1ff92d0fc1/src/browser/Navigators/Navigator.js#L34-L41) is the type of the `flags` argument that needs to be passed to `Deck.open`. Inlining here for convenience:

```js
export type Flags =
  { output: Output.Flags
  , isPinned?: boolean
  , isInputEmbedded?: boolean
  , input?: Input.Flags
  , overlay?: Overlay.Flags
  , assistant?: Assistant.Flags
  }
```

As of `output` field it's type is defined in `WebView` [as `Flags` type](https://github.com/linclark/browserhtml/blob/a62789ecb28a310e51d447828bf22e1ff92d0fc1/src/browser/Navigators/Navigator/WebView.js#L49-L57) inlining it here for connivence:

```js
export type Flags =
  { uri: URI
  , disposition: Disposition
  , name: string
  , features: string
  , options?: {}
  , ref: any
  , guestInstanceId: ?string
  }
```

In this instance `name` field is missing in your change, which is what flow is attempting to tell in the following error:

```
src/browser/Navigators.js:101
  Member 1:
   29:   | { type: "Open", open: flags }
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ object type
  Error:
   35:   { output: Output.Flags
                   ^^^^^^^^^^^^ property `name`. Property not found in. See: src/browser/Navigators/Navigator.js:35
  188:             { uri: Runtime.env.url
                   ^ object literal. See: src/browser/Navigators.js:188
```

Rest of the error from flow is I believe are just a noise caused by this type mismatch.

linclark:
Thanks for walking through the types with me, that was very helpful.

It seems like I also needed to explicitly typecast `Runtime.env.url` to string. It's type is defined as the union of string and array, so I was getting this error:

```
  Error:
  188:             { uri: Runtime.env.url
                          ^^^^^^^^^^^^^^^ array type. This type is incompatible with. See: src/browser/Navigators.js:188
   50:   { uri: URI
                ^^^ string. See: src/browser/Navigators/Navigator/WebView.js:50
```

I also decided to wrap it in `URI.read` so that the http scheme would be added if the user omitted the scheme.

Gozala:
> Thanks for walking through the types with me, that was very helpful.
> 

Great, I was little afraid to not over do it :)

> 
> It seems like I also needed to explicitly typecast Runtime.env.url to string. It's type is defined as the union of string and array, so I was getting this error:

Right, I forgot about that. As technically you could do `index.html?url=foo.com&url=bar.com` in which case you'll end up with `Runtime.env.url` being array of strings. In such case type casting via `String(Runtime.env.url)` is not a great option because you wind up with `'foo.com,bar.com'`. Instead it would make sense to open separate tab for each url or just use either first or last from array (although in that case type checker will force you into checking that array isn't empty).

Anyway I think this is good enough for now, followup changes could add support for handling multiple url passing.

> 
>   Error:
>   188:             { uri: Runtime.env.url
>                           ^^^^^^^^^^^^^^^ array type. This type is incompatible with. See: src/browser/Navigators.js:188
>    50:   { uri: URI
>                 ^^^ string. See: src/browser/Navigators/Navigator/WebView.js:50
> I also decided to wrap it in URI.read so that the http scheme would be added if the user omitted the scheme.

Cool that makes sense!

Gozala:
@linclark Thanks!

P.S.: If you'll find time to document this new option in a wiki that would be really great:
https://github.com/browserhtml/browserhtml/wiki/Browser-options

PR #1194, opened 2016-07-28
Factor out services form the History and Search modules.

Gozala:
## This is part of the #1185 and is based on top of #1195

This change factors out IO stuff from `Search` and `History` modules into separate service modules.

### DIFF
--- a/src/Service/Search.js
+++ b/src/Service/Search.js
@@ -1,0 +1,100 @@
+/* @flow */
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import {Task} from "reflex";
+import {ok, error} from "../common/result";
+import * as Search from "../browser/Navigators/Navigator/Assistant/Search"
+
+const pendingRequests = Object.create(null);
+
+export const query =
+  (query:string, limit:number=100):Task<Error, Array<Search.Match>> =>
+  new Task((succeed, fail) => {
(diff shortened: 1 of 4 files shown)


src/Service/Search.js
```diff
@@ -0,0 +1,100 @@
...
+    const request = new XMLHttpRequest({ mozSystem: true })
+    pendingRequests[query] = request
+    const url = `https://ac.duckduckgo.com/ac/?q=${query}&type=list`
+    request.open('GET', url, true)
+    request.responseType = 'json'
+
+    request.onerror = event => {
+      delete pendingRequests[query]
+      fail(Error(`Network request to ${url} has failed: ${request.statusText}`))
+    }
```
paulrouget: Probably want to add `onabort` here…

src/Service/Search.js
```diff
@@ -0,0 +1,100 @@
...
+    request.send()
+  })
+
+export const abort = <never>
+  (query:string):Task<Error, never> =>
+  new Task((succeed, fail) => {
+    const request = pendingRequests[query]
+    if (request != null) {
+      request.abort()
+      delete pendingRequests[query]
```
paulrouget: … and not call `delete pendingRequests[query]`.

src/Service/Search.js
```diff
@@ -0,0 +1,100 @@
...
+
+const requestURL =
+  request =>
+  request.requestURL ||
+  request.url ||
+  crash(Error('Can not find request url'))
+
+const fail =
+  request =>
+  error(Error('Can not decode ${JSON.stringify(request.response))} from ${requestURL(request)} '))
```
paulrouget: backquote?

paulrouget:
r=me

PR #1195, opened 2016-07-28
Rename assistant modules

Gozala:
## This is part of the larger #1185 change

On it's own this change seems stupid but it's easier to do factor out such tiny changes and land them incrementally then whole thing where things make sense but is really difficult to see whole change.


paulrouget:
r=me

PR #1196, opened 2016-07-28
Implement appendFX helper function.

Gozala:
## This is part of the large #1185 change.

This utility helps including extra effects in the update step

```js
appendFX(Effects.perform(task), updateAThink(state, action))
```


paulrouget:
r=me

PR #1197, opened 2016-07-28
Address port

Gozala:
## This is part of the #1185 and is based on top of #1196

Utility `port` function allows creation of endpoints to which `address` could be passed.

```js
export const onFocus = port(always({ type: “Focus” }))
html.input({ onFocus: onFocus(address) })
```

Idea is that `Port` instances could be shared and reused across modules.


paulrouget:
r=me

PR #1198, opened 2016-07-28
Address port anotate

Gozala:
## This is part of the #1185 and is based on top of #1197

Utility `anotate` function allows mapping ports such that resulting port will map it's messages via provided `tag` function:

### `Focus.js`
```js
export const onFocus = port(always({ type: “Focus” }))
```

### `Input.js`

```js
import * as Focus from "./Focus"

const tagFocus = message => ({ type: "Focus", focus: message });

// P.S.: Same `tagFocus` is used with `fx.map(tagFx)` 
const onFocus = anotate(Focus.onFocus, tagFocus);

const view = (model, address) =>
  html.input({ onFocus: onFocus(address) /*...*/ })

const update = (model, message) => {
  switch (message.type) {
    case "Focus":
      return setFocus(model, Focus.update(model.focus, message.focus));
      /* ... */
  }
}
```

### DIFF
--- a/src/common/prelude.js
+++ b/src/common/prelude.js
@@ -206,6 +206,14 @@ type Port <event, message> =
   (address:Address<message>) =>
   Address<event>
 
+
+export const anotate = <tagged, message, event>
+  ( port: Port<event, message>
+  , tag: (input:message) => tagged
+  ):Port<event, tagged> =>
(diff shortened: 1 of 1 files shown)


paulrouget:
r=me

PR #1199, opened 2016-07-29
Refactor suggestion

Gozala:
## This is part of the large #1185 change & based off #1194

This moves part of the view logic from the #1185 change and encapsulates in last three commits. Other noise is due to changes from #1194


paulrouget:
r=me

PR #1200, opened 2016-07-29
Make Editable and Focusable embeddable.

Gozala:
## This is part of the larger #1185 change

Editable and Focusable are implemented as interfaces or components that you need to mix in. This a step towards making them more standard components that can be embedded pretty much like all other components are.


paulrouget:
r=me

PR #1201, opened 2016-07-29
Refactor `TextInput` to avoid mixins.

Gozala:
## This is part of the larger #1185 change & is based off #1200

Refactor TextInput to embed `Focusable`, `Editable`, `Control` instead of mixing them in.


paulrouget:
r=me

PR #1202, opened 2016-07-29
Refactor `Input` to avoid mixins.

Gozala:
## This is part of the larger #1185 change & is based off #1201

Refactor `Input` to embed `Focusable`, `Editable` instead of mixing them in.


paulrouget:
r=me

PR #1203, opened 2016-07-29
Simple toggle

Gozala:
## This is part of the larger #1185 change & is based off #1202

Refactor `Toggle` to embed `Button` and avoid mixins.

PR #1204, opened 2016-07-29
Refactor `Button` to avoid mixins

Gozala:
## This is part of the larger #1185 change & is based off #1203

Embed `Control`, `Target` & `Focus` modules instead of mixing them in.


paulrouget:
r=me

PR #1205, opened 2016-07-29
Make `Focus` implementation monomorphic.

Gozala:
## This is part of the larger #1185 change & is based off #1204

Make `Focus` implementation monomorphic.

### DIFF
--- a/src/common/toggle.js
+++ b/src/common/toggle.js
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 
-import {merge, always, nofx, mapFX} from "../common/prelude"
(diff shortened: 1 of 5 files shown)


src/common/toggle.js
```diff
@@ -5,7 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 
-import {merge, always, nofx, mapFX} from "../common/prelude"
+import {tag, tagged, anotate, nofx, mapFX, always} from '../common/prelude';
```
paulrouget: Is all of that needed?

paulrouget:
r=me

src/common/toggle.js
Gozala: > Is all of that needed?

Let's deal with all the redundant imports as part of #1097

PR #1208, opened 2016-07-29
Make `Editable` implementation monomorphic.

Gozala:
## This is part of the larger #1185 change & is based off #1207

Make `Editable` implementation monomorphic.


paulrouget:
r=me

PR #1213, opened 2016-08-03
Use iframe.setVisibile on mozbrowser iframes

Gozala:
Set visibility on/of if frame is / isn’t selected.

First commit also adds a generic hook into virtual-dom, that runs a task with corresponding DOM element and passed value to do custom modifications when patch is applied.

I have being testing this feature via command:

```
/Applications/Servo.app/Contents/MacOS/servo -w -b --pref dom.mozbrowser.enabled --pref dom.forcetouch.enabled --pref  shell.builtin-key-shortcuts.enabled=false http://localhost:6060\?url\=https://Gozala.jsbin.com/mudum
```

Open tab increments number on animation frames, so if you switch to other tab it will stop incrementing & resume when you're back.

P.S.: I will do a followup to use same `setting` hook for other workarounds we currently have.

### DIFF
--- a/src/browser/Navigators/Navigator.js
+++ b/src/browser/Navigators/Navigator.js
@@ -910,7 +910,7 @@ export const render =
       , styleBackground(model.output)
       )
     }
-  , [ Output.view(model.output, forward(address, tagOutput))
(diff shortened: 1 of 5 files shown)


paulrouget:
r=me

PR #1214, opened 2016-08-03
Provide type annotations for @driver module

Gozala:
This change follows up on #1213 so please ignore first two commits. It also does two changes:

1. Creates `index.js.flow` so flow can ensure that `@driver` module is used properly.
2. Special `Abort` type is replaced with `null` which aborts feeding event back into update loop. This change was done as otherwise type annotations were becoming too complex and was not worth it.


paulrouget:
r=me

PR #1215, opened 2016-08-03
Deploy on each revision

paulrouget:
I haven't tested, but that should do it.

Once this is done, we'll update Servo to not use a numbered revision but the head of the crate branch.

### DIFF
--- a/.bin/deploy.sh
+++ b/.bin/deploy.sh
@@ -5,32 +5,51 @@ set -e
 git config --global user.email "travis@browser.html"
 git config --global user.name "travis"
 
-TAG=`git describe --abbrev=0 --tags`
-LATEST_REV=`git rev-parse $TAG`
+REMOTE_MASTER_HEAD=`git ls-remote origin master | cut -f 1 -d$'\t'`
 REPO_URL=`git config --get remote.origin.url`
 HEAD=`git rev-parse HEAD`
-DEPLOYED_REV=`curl --silent https://raw.githubusercontent.com/browserhtml/browserhtml/gh-pages/HEAD`
+LATEST_TAG=`git describe --abbrev=0 --tags`
+REV_FOR_LATEST_TAG=`git rev-parse $LATEST_TAG`
 
-if [ "$HEAD" = "$LATEST_REV" ]; then
-  if [ "$HEAD" != "$DEPLOYED_REV" ]; then
-    echo "Deploying version ${TAG}"
-    rm -rf dist
-    mkdir dist
-    git clone -b gh-pages $REPO_URL dist
-    rm -r dist/*
-    npm run build
-    npm run build-cargo
-    cd dist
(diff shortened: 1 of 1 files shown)


.bin/deploy.sh
```diff
@@ -5,32 +5,21 @@ set -e
...
+rm -r dist/*
+npm run build
+npm run build-cargo
+cd dist
+echo "" > .nojekyll
+git add -A
+git commit -m "Deploy version ${HEAD}"
+git push --quiet "https://travis:${GH_TOKEN}@${GH_REF}" ${$DEPLOY_BRANCH} > /dev/null 2>&1
+git status
+echo "Deployed version ${TAG}"
```
Gozala: You have removed `TAG` variable so maybe you want to echo `"Deployed ${HEAD}"` instead ?

.bin/deploy.sh
```diff
@@ -5,32 +5,21 @@ set -e
...
+rm -rf dist
+mkdir dist
+git clone -b ${DEPLOY_BRANCH} ${REPO_URL} dist
+rm -r dist/*
+npm run build
+npm run build-cargo
+cd dist
+echo "" > .nojekyll
+git add -A
+git commit -m "Deploy version ${HEAD}"
```
Gozala: I think `"Deploy ${HEAD}"` makes more sense.

.bin/deploy.sh
```diff
@@ -5,32 +5,21 @@ set -e
...
 git config --global user.name "travis"
 
-TAG=`git describe --abbrev=0 --tags`
-LATEST_REV=`git rev-parse $TAG`
 REPO_URL=`git config --get remote.origin.url`
 HEAD=`git rev-parse HEAD`
-DEPLOYED_REV=`curl --silent https://raw.githubusercontent.com/browserhtml/browserhtml/gh-pages/HEAD`
+DEPLOY_BRANCH="crate"
 
-if [ "$HEAD" = "$LATEST_REV" ]; then
```
Gozala: You still need to check that this is running on master branch and do not deploy otherwise or we'll end up deploying on every pull request which is not what we want.

Gozala:
@paulrouget can we still keep deploy on tags and push to `gh-pages` ? That known stable version will still be available.

.bin/deploy.sh
paulrouget: > You still need to check that this is running on master branch and do not deploy otherwise or we'll end up deploying on every pull request which is not what we want.

When can this happen? Travis will only call deploy.sh on master when there's a new set of commits, no?

paulrouget:
@gozala now it deploys into a "crate" branch on every new commit to master, and to "gh-pages" on every new tagged commit.

Gozala:
R+

I apologize for misleading, you were right according to .travis.yml it would run deploy script only on master snd tags

PR #1218, opened 2016-08-03
Use xcopy on windows

vvuk:
On windows, if we're building without a full msys-like environment, we can't do cp -r.  This has build.rs use xcopy in this case.


tschneidereit:
_Sigh_ indeed.

PR #1219, opened 2016-08-03
Report non-pipeline panics

### LINKED ISSUE #1182: Handle non-pipeline thread panics in the root document.
Servo will dispatch mozbrowsererror's on the root document for non-pipeline thread panics. We should handle them and hook them into existing error reporter code base.

Gozala:
Fixes #1182

### DIFF
--- a/src/browser.js
+++ b/src/browser.js
@@ -85,6 +85,7 @@ export type Action =
   | { type: "Modify", modify: ID, action: any }
   | { type: "Open" }
   | { type: "Tabs", tabs: Tabs.Action }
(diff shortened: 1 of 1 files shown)


paulrouget:
Is it possible to test a non-pipeline panic?

paulrouget:
Looks good to me, but I'd like to test.

paulrouget:
Just saw https://github.com/servo/servo/issues/12727

Gozala:
Let's land, even though we can't test cause in worst case it just won't report issues in best case it will work as expected ;)

PR #1222, opened 2016-08-05
zoom in/out animation with css transitions

### LINKED ISSUE #1220: Implement zoom in / out animation with a CSS transition
Got blocked by https://github.com/servo/servo/issues/12749

Gozala:
# Do not land, blocked by https://github.com/servo/servo/issues/12749

Fixes #1220

### DIFF
--- a/src/browser/Navigators.js
+++ b/src/browser/Navigators.js
@@ -1,7 +1,6 @@
 /* @flow */
 
 import * as Deck from "../common/Deck"
-import * as Animation from "../common/Animation"
 import * as Unknown from "../common/unknown"
 import * as Display from "./Navigators/Display"
 import {Effects, html, forward, thunk} from "reflex"
@@ -39,7 +38,6 @@ export type Action =
   | { type: "SelectNext" }
   | { type: "SelectPrevious" }
   | { type: "Crash", crash: Report }
-  | { type: "Animation", animation: Animation.Action }
   | { type: "Tabs", tabs: Tabs.Action }
   | { type: "Deck", deck: Deck.Action<Navigator.Action, Navigator.Flags> }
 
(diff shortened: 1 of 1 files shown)


paulrouget:
I believe this is better because for performance, but can you explain why we want to do that now?

tschneidereit:
> I believe this is better because for performance, but can you explain why we want to do that now?

We had talked about this in May/June and postponed it until after the demo and Nightly. @pcwalton thought it's quite important to do this, and I can easily believe that to be true.

@Gozala, looks like the blocker has been fixed, so I guess we can go forward with this now?

src/browser/Navigators.js
```diff
@@ -507,6 +427,22 @@ const styleSheet = Style.createSheet
       , overflow: 'hidden'
       , transformOrigin: 'left center'
       , boxSizing: 'border-box'
+      , transition: `border-right 200ms ${easeOutCubic}, transform 200ms ${easeOutCubic}`
+      , background: 'white'
+      }
+    , shrinked: {
```
tschneidereit: To be proper English, this would need to be `shrunken`

src/browser/Navigators.js
```diff
@@ -507,6 +427,22 @@ const styleSheet = Style.createSheet
...
+        borderRight: 'solid transparent 50px'
+      }
+    , expanded: {
+        borderRight: 'solid transparent 0px'
+      }
+    , collapsed:
+      { transform: 'translate3d(0, 0, 0)'
+      , transition: `border-right 200ms ${easeOutCubic}, transform 200ms ${easeOutCubic}`
```
tschneidereit: Won't this be copied from `base` and thus already exist?

src/browser/Navigators.js
```diff
@@ -507,6 +427,22 @@ const styleSheet = Style.createSheet
...
+      }
+    , expose:
+      { transition: `border-right 200ms ${easeOutCubic}, transform 500ms ${easeOutCubic}`
```
tschneidereit: Same here.

Gozala:
Closing this for now.

PR #1226, opened 2016-08-11
Use xcopy for build.rs on windows

vvuk:


tschneidereit:
I should've seen that #1218 merged into the wrong branch, which is automatically overwritten upon checkin to `master`, sorry.

r=me

PR #1228, opened 2016-08-11
Workaround iframe API limitations at the DOM level.

### LINKED ISSUE #949: servo+browser.html - entering an url in the url bar doesn't seem to work after navigating a page
Not sure whether it belongs to servo or browser.html. On a local build of [today's servo](https://github.com/servo/servo/commit/4341dc5fb45a3e4d5397ca64e7aabe85f8ac23ab) on Linux:

1) ./mach run --release -w -b
2) enter reddit.com/r/rust in the url bar and press enter
3) visit a link from the reddit page
4) enter reddit.com/r/rust again in the url bar
5) nothing happens

Gozala:
We used to keep track `src` and currently loaded `url` as separate fields in the model so that we would only update `src` when user edited url. But that logic had a major issue #949:

If user initial loads url `a` than navigates to `b` by clicking a link and then edits url to navigate to `a` nothing would happen. That is because Virtual DOM would diff and see that src in both cases is `a` and do not update `iframe.src`.

With this change we use `setting(location, url)` for updating `iframe.src`. As of `location` setter it keeps track of iframe location and only reflects passed `url` if it is different from it.

This approach fixes #949 and makes state model simpler.

### DIFF
--- a/src/browser/Navigators/Navigator.js
+++ b/src/browser/Navigators/Navigator.js
@@ -694,11 +694,11 @@ const editInput =
   ( update
   , model
   , [ ActivateInput
-      // @TODO: Do not use `model.output.navigation.currentURI` as it ties it
+      // @TODO: Do not use `model.output.navigation.url` as it ties it
       // to webView API too much.
     , ( model.isInputEmbedded
       ? SetSelectedInputValue('')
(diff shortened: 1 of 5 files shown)


paulrouget:
It appears to work as expected.

r=me

PR #1229, opened 2016-08-12
Subtract backtrace when issue URL is over the limit.

### LINKED ISSUE #1165: Bug Reports via GET too long for NGINX, 414, should probably be a POST
Attempting to play a video on YouTube caused an error, a report button was displayed with a stacktrace in the app, but when I clicked on it, it tried to submit the bug, but the URI was too long on the GET to NGINX, so NGINX rejected it with a 414 code.

This should probably be marshaled to something and sent to a POST endpoint instead.

Gozala:
With this change crash report backtrace will be subtract by number of characters that went over the URL limits.

Fixes #1165


src/browser/IssueReporter.js
```diff
@@ -78,21 +78,36 @@ const submit =
...
+    const report = `https://github.com/servo/servo/issues/new?title=${encodeURIComponent(description)}&body=${body}`;
+    // If report URL exceeds subtract backtrace to stay with in the limits.
+    const result =
+      ( report.length <= URL_LENGTH_LIMIT
+      ? report
+      : encodeReport
+        ( { description
+          , url
+          , version
+          , backtrace: backtrace.substr(0, report.length - URL_LENGTH_LIMIT - 4) + '\n...'
```
jsantell: What is `4`? Should be a constant I'd imagine.

jsantell:
lgtm

src/browser/IssueReporter.js
Gozala: > What is 4? Should be a constant I'd imagine.

Oh it's for the added `\n...` there

PR #1231, opened 2016-08-13
Use xcopy only with msvc target, as dir paths are different with mingw

vvuk:
This needs to just be for msvc, as for mingw dir paths get specified using mingw syntax. Sigh.


paulrouget:
r=me

PR #1233, opened 2016-08-15
disable sandboxing and multiprocess

paulrouget:


Gozala:
@paulrouget do we have any issues with sandboxing as well ? Multiprocess does cause panics but as far as I could tell sandboxing was fine.

Gozala:
Ok just found servo/servo#12791 so that answers my question.

paulrouget:
> do we have any issues with sandboxing as well ? Multiprocess does cause panics but as far as I could tell sandboxing was fine.

Also, sandboxing can't work without multiprocess.

PR #1234, opened 2016-08-15
Implement fallback icons

### LINKED ISSUE #973: Use a fallback favicon, the same way Safari does
Go to example.com. There's no favicon defined. Now, pin example.com in Safari. They build a favicon based on the first letter of the title.

Gozala:
## Fixes #973

Now with pages that don't have favicon you'll see generated icon that has background of a site `theme-color` (see [mozbrowsermetachange](https://developer.mozilla.org/en-US/docs/Web/Events/mozbrowsermetachange) event for more details) and a contrasting foreground color.

Unfortunately [`mozbrowsermetachnage`](https://developer.mozilla.org/en-US/docs/Web/Events/mozbrowsermetachange) events don't seem to be dispatched in servo so colors default dark gray over warm white. Once events will be dispatched this code will take advantage of it.

Unfortunately due to https://github.com/servo/servo/issues/12885 `onerror` events on images aren't fired in servo there for in some cases tabs will have no fallback icons, as we will try to load unavailable image but will never discover it was not available there for won't fallback to generated icon.

I have being testing this with following pages:
https://Gozala.jsbin.com/yugoqil
https://Gozala.jsbin.com/zabifib
http://fitzgeraldnick.com

And observe following results:

## Electron

screen shot 2016-08-15 at 4 58 08 pm

## Servo

screen shot 2016-08-15 at 5 00 35 pm

Differences are due to issues described above + electron seems to fire `iconchange` event regardless of it being present or not, hence `jsbin` icon in electron version.

### DIFF
--- a/src/browser/Navigators/Navigator/WebView.js
+++ b/src/browser/Navigators/Navigator/WebView.js
@@ -25,11 +25,11 @@ import * as MozBrowserFrame from './WebView/MozBrowserFrame';
 import * as ElectronFrame from './WebView/ElectronFrame';
 import * as Ref from '../../../common/ref';
 import * as Tab from '../../Sidebar/Tab';
+import * as Favicon from '../../../common/favicon';
 
 
 import type {Address, DOM} from "reflex"
 import type {URI, Time, Integer, Float} from "../../../common/prelude"
-import type {Icon} from "../../../common/favicon"
 import {performance} from "../../../common/performance"
 import type {Report} from "../../IssueReporter"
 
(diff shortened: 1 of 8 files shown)


src/browser/Navigators/Navigator/WebView.js
```diff
@@ -286,6 +288,17 @@ const updatePage = cursor
...
   );
 
+const delegateTabUpdate =
+  (model, action) => {
+    switch (action.type) {
+      case "Page":
+        return updatePage(model, action.page)
+      default:
+        return updateTab(model, action)
+
```
tschneidereit: nit: remove blank line.

src/browser/Navigators/Navigator/WebView/Page.js
```diff
@@ -161,23 +231,50 @@ const updateMeta =
...
+    : Pallet.blank
+    )
+  );
+
+const chooseInital =
+  (hostname:string, title:string) =>
+  ( hostname == ""
+  ? title
+  : hostname
+  ).substr(0, 1).toUpperCase()
```
tschneidereit: Nit: use `charAt(0)` here instead of `substr`.

tschneidereit:
r=me

tschneidereit:
(With feedback addressed, that is.)

PR #1235, opened 2016-08-15
Improve favicon selection logic

Gozala:
## Is based on top of #1234 

Electron ends up having better favicons due to ignoring apple-touch stuff and in most cases that was due to `rel=icon` vs expected `rel=shorctur icon` so this patch handles both.

### DIFF
--- a/src/common/favicon.js
+++ b/src/common/favicon.js
@@ -9,14 +9,23 @@ import type {URI} from "../common/prelude"
 
 export type {URI}
 
-export type Icon =
+export type Model =
   { href: URI
   , sizes?: string
   , rel: ?string
   }
 
 
-const constructFaviconURI = (href, size) => `${href}#-moz-resolution=${size},${size}`;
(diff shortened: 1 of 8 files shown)


src/common/favicon.js
```diff
@@ -48,9 +48,9 @@ export const getBestIcon =
   const others = new Set();   // store icons without sizes or non-shortcut icons
 
   for (var icon of icons) {
-    if (icon.rel != 'shortcut icon') {
+    if (icon.rel != 'shortcut icon' && icon.rel != 'icon') {
       others.add(icon);
-    } else if (!icon.sizes) {
+    } else if (!icon.sizes || icon.sizes === "") {
```
tschneidereit: This shouldn't be required: empty strings are falsy, so this should be functionally identical to what you had before.

tschneidereit:
r=me with superfluous change reverted.

PR #1241, opened 2016-08-19
Setup standard to enforce style on new code

### LINKED ISSUE #1240: Setup standardjs to enforce it's style on new files
Nobel goal #1238 of reformatting all of the existing code to comply  with standard is really cool, but in the meantime it would make sense to enforce standard at least on new files. This is what this issue is about.

Gozala:
Fixes #1240

This change configures eslint on top of [eslint-config-standard](https://github.com/feross/eslint-config-standard) which is then configured to workaround following issues preventing us using from stock standard package:

1. space-infix-ops reports polymorphic functions:
   https://github.com/babel/babel-eslint/issues/366
2. no-duplicate-imports rule triggered by `import type`:
   https://github.com/feross/standard/issues/599

All of the currently tracked fils are also listed under `.eslintignore` so that current code will not be listed until we get to reformat it.


tschneidereit:
I'm fine with this overall, except for one major thing: I'm very strongly opposed to the "no semicolos" rule. The vast majority of existing code does use semicolons, meaning that people (including myself) use it as a visual aide when scanning code. It might well be that there are no _technical_ reasons for not using semicolons as long as you adhere to a bunch of other rules. The same is true for, e.g., almost all whitespace, though, and doesn't mean it's a good idea.

I know that there's a substantial set of people preferring not to use semicolons, but from what I can tell they are still a significant minority.

FWIW, I generally vastly prefer a "more traditional" style such as [airbnb's](https://github.com/airbnb/javascript), but given how fruitless style discussions are, let's not engage in them where it's not needed. The semicolons thing I think *is* an important decision.

paulrouget:
I agree with Till.

Gozala:
After discussing this in the meeting we have decided to give it a try & maybe review this decision in the future.

Issue #1242, opened 2016-08-19
Reformat code to follow standard


Gozala:
Ok so once #1241 lands we'll be linting all new code to comply with standard style. Unfortunately automatically reformatting all of the existing code tracked by #1238 seems far from trivial task and still unclear if it's a worthy effort.

Alternative approach would be to just run a lint file by file and address errors programmatically or manually. As far as I can tell `eslint --fix ./path/to/file.js` seems to address majority of errors, but not the comma first issues (see https://github.com/eslint/eslint/issues/6941).

This is also maybe an easy first contribution patch. In case you want to contribute to this issue you should:
1. Remove path for the file you're going to fix from `.eslintignore` file.
2. Run lint and address it's errors.
3. Submit a change via pull request.

Note: This way we can gradually reformat our code.

P.S: No need to reformat stuff in `src/About/Newtab/Demos/` as that code is not maintained in this repo.

onurtemizkan:
Hey @Gozala , I'm going to work on this. But just to be sure, you prefer one PR for each file, right? It would be a lot easier if I submit a PR for each directory of files. What do you think?

Gozala:
> Hey @Gozala , I'm going to work on this. 

Thanks that would really great!

> But just to be sure, you prefer one PR for each file, right? 

If it's just formatting changes (as it should be) I don't mind multiple files or even all files in one PR.

> It would be a lot easier if I submit a PR for each directory of files. What do you think?

Only reason I think file by file PRs would be better though is because that would reduce possibility of merge conflicts.

magopian:
The command isn't `eslint` anymore, but `standard-flow`, and I believe the process is now as follows:
- remove the file you want to work on from the `standard-flow: { "ignore" : ...}` list in the `package.json file`
- run `npm run lint` (or `node_modules/.bin/standard-flow --fix` to have some of the warnings automatically fixed)
- fix the (remaining) errors manually
- finally run `npm run test` to make sure nothing's broken
- rinse and repeat

PR #1245, opened 2016-09-03
Reformat Demos to StandardJS style (#1242)

onurtemizkan:
(refs: #1242)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,9 +68,6 @@
   },
   "standard-flow": {
     "ignore": [
-      "src/About/Newtab/Demos/evil_ad/physics/physics.js",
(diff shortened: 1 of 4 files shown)


Gozala:
Oh I am so sorry should have said it in #1242 that we don't really need to reformat stuff in `src/About/Newtab/Demos/` as that code is not really maintained by us.

Gozala:
@onurtemizkan I have merged this anyway but it could be we'll replace this demo with new code at some point that won't follow standard. Please do not reformat other demos as their source isn't really maintained here it was just easiest way bundle servo demos with nightlies.

PR #1246, opened 2016-09-05
Reformat [About/Newtab] to StandardJS Style (#1242)

onurtemizkan:
refs: #1242

### DIFF
--- a/src/About/Newtab/Newtab.js
+++ b/src/About/Newtab/Newtab.js
@@ -4,140 +4,129 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import {merge, tag, tagged} from "../../Common/Prelude"
-import {Effects, html, thunk, forward} from "reflex"
-import {Style, StyleSheet} from "../../Common/Style";
-import {cursor} from "../../Common/Cursor";
-import * as Unknown from "../../Common/Unknown";
+import {merge} from '../../Common/Prelude'
+import {Effects, html, thunk, forward} from 'reflex'
+import {Style, StyleSheet} from '../../Common/Style'
+import {cursor} from '../../Common/Cursor'
+import * as Unknown from '../../Common/Unknown'
 
-import * as Tiles from './Newtab/Tiles';
-import * as Wallpapers from './Newtab/Wallpapers';
+import * as Tiles from './Newtab/Tiles'
+import * as Wallpapers from './Newtab/Wallpapers'
 
+import type {Address, DOM} from 'reflex'
 
-import type {Address, DOM} from "reflex"
+export type Model = {
+  wallpapers: Wallpapers.Model,
+  tiles: Tiles.Model
+}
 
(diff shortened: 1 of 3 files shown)


Gozala:
Hi @onurtemizkan did you automate this process somehow ? I think some of the formatting looks really hard to read (which is why we did not go after automated solution actually). I will suggest few changes inline, I hope you'll understand.

src/About/Newtab/Newtab.js
```diff
@@ -4,140 +4,129 @@
...
+  tiles: Tiles.Model
+}
 
+export type Action = | {
+      type: "Wallpapers",
+      wallpapers: Wallpapers.Action
+    } | {
+      type: "Tiles",
+      tiles: Tiles.Action
+    }
```
Gozala: Can we please go with following instead:

```js
export type Action =
  | { type: "Wallpapers", wallpapers: Wallpapers.Action }
  | { type: "Tiles", tiles: Tiles.Action }
```

src/About/Newtab/Newtab.js
```diff
@@ -4,140 +4,129 @@
...
+  update: Tiles.update,
+  tag: TilesAction
+})
+
+export const update = (model:Model, action:Action):[Model, Effects<Action>] => (
+  action.type === 'Wallpapers' ? updateWallpapers(model, action.wallpapers)
+  : action.type === 'Tiles' ? updateTiles(model, action.tiles)
   : Unknown.update(model, action)
-  );
+)
```
Gozala: Hmm. I believe according to standardjs it should be:

```js
( action.type === 'Wallpapers'
? updateWallpapers(model, action.wallpapers)
: action.type === 'Tiles'
? updateTiles(model, action.tiles)
: Unknown.update(model, action)
)
```

Actually if you don't mind doing it would be great to just rewrite our updates to `switch` statements (that we kind of started doing already here and there) as it is more readable and actually executes faster:

```js
switch (action.type) {
  case 'Wallpapers':
    return updateWallpapers(model, action.wallpapers)
  case 'Tiles':
    return updateWallpapers(model, action.wallpapers)
  default:
    return Unknown.update(model, action)
}
```

src/About/Newtab/Newtab.js
```diff
@@ -4,140 +4,129 @@
...
-      , style: Style
-        ( styleSheet.base
-        , readWallpaper(activeWallpaper)
-        )
-      }
-    , [ html.meta
-        ( { name: "theme-color"
-          , content: activeWallpaper.color
+    html.div(
+      {
```
Gozala: I think this should be formatted like:

```js
return html.div({
  // ...
}, [
  // ...
])
```

onurtemizkan:
> Hi @onurtemizkan did you automate this process somehow ? I think some of the formatting looks really hard to read (which is why we did not go after automated solution actually). I will suggest few changes inline, I hope you'll understand.

Hi @Gozala. No I'm just reformatting manually but sometimes it becomes hard to choose the best legible option, so please continue to make suggestions in my future PR's too.

PR #1254, opened 2016-09-28
Reformat [About/Repl/] to StandardJS Style (#1242)

onurtemizkan:
refs: #1242


src/About/Repl/Repl.js
```diff
@@ -4,250 +4,224 @@
...
-    , evaluate: action.submit
+const CreateCell = ({
+  type: 'CreateCell'
+})
+
+const Focus = ({
+  type: 'Focus'
+})
+
+const Evaluate = (id, input) => ({ // eslint-disable-line no-unused-vars
```
onurtemizkan: This function is unused I'm just disabling rule for now.

PR #1263, opened 2017-01-12
Update path to a js bundle from an html file.

Gozala:
Fix for servo/servo#14919

---
This change is [Reviewable](https://reviewable.io/reviews/browserhtml/browserhtml/1263)

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,5 +20,5 @@
   </head>
   <body tabindex="-1">
   </body>
-  <script src="components/main.js"></script>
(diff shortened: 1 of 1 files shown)


Gozala:
@paulrouget could you also please check that fix actually resolved the issue and there's nothing else is broken on Linux. It's just setting up a Linux VM and running on this tiny laptop would take me a lot of time.

paulrouget:
It works.

PR #1272, opened 2017-03-13
Update fontawesome woff for Windows

jonathandturner:
This fixes https://github.com/servo/servo/issues/15255 on my box.  I pulled down FontAwesome's ttf and reconverted it to woff.  The result is a little bigger than what's there now, so perhaps there were missing sigils?


paulrouget:
Thanks! This will require a cargo update in servo.

PR #1275, opened 2017-04-05
Remove outdated & obsolete docs.

Gozala:
Fixes #447


tschneidereit:
Thank you, much better to remove these docs than to leave misleading documentation around.

PR #1277, opened 2017-04-07
Electron runtime improvements

Gozala:
## Overview of changes

- Only open devtools if when `--devtools` flag is passed.
- Wire up window minimize and maximize actions.
- Differentiate between window close and app quit actions.
- Persist logins across browsing sessions.
- Use browser.html icon instead of electron default.


tschneidereit:
LGTM

PR #1278, opened 2017-04-07
Fix regression in button code

Gozala:
Regression was introduced by the last code refactoring that did not properly wired some button actions & styling.


tschneidereit:
Looks like this was all contained in the last two PRs I just merged, so I'll close this. Please reopen if I'm missing something.

PR #1296, opened 2017-05-06
Readme fixes

atouchet:

PR #1300, opened 2017-06-02
Fix links in README.md

jdanford:


jdanford:
Whoops, didn't realize #1296 already existed! This is essentially a superset of that PR, so if the additional changes are alright then whoever merges this should close that at the same time. Maybe @Gozala?

Gozala:
Thanks!

Issue #1264, opened 2017-01-26
New tab and tab list buttons have small hitboxes


fvsch:
With Servo’s latest nightly on macOS, the "new tab" and "tab list" buttons are really small and close together. I don’t know how to inspect the precise element area, but just using the mouse cursor I see them as just taking the width and height of the icon.

A more reasonable UI would have the buttons further apart and each having a bigger reactive area, around 30x30 points or so.

![servo-browserhtml-buttons](https://cloud.githubusercontent.com/assets/243601/22326742/37470470-e3b4-11e6-8a4b-c0e8b9034d36.png)

Looking at some images in browserhtml’s repo and wiki, some show small buttons, and other show buttons with more breathing room, but they might be mockups only.

Is this usability issue fixed or improved upon already, but perhaps not uplifted to Servo?

gordonbrander:
I thought we had made the hit boxes larger as illustrated. A _slim_ possibility this could be a Servo bug? Taking a look with the remote inspector would answer this.

joshua-s:
This was a current issue and I fixed it in my PR :)

PR #1306, opened 2017-10-05
Expand header button hitboxes

joshua-s:
Fixes https://github.com/browserhtml/browserhtml/issues/1264 by giving the header buttons hitboxes of 30px by 30px.


paulrouget:
Thank you.

